PR #212, opened 2014-11-15
[WIP] Basic slope support

eevee:
I like slopes, which has made 3D mode a little less than useful, so I took a crack at this.

It even correctly avoids warning about missing textures in basin-style architecture!

Only understands `Plane_Align` and tilt things at the moment, because those are all I'm using at the moment.  Also has a lot of TODOs.  And could probably stand to compute the floor/ceiling planes _once_ rather than over and over all the time.  And some other things.  But it's a start.

I went off the beaten path a bit and wrote the slope-finding function as a `template<PlaneType>`, where `PlaneType` is just an enum of floor and ceiling.  A couple similarly-templated helpers took care of the details, and I only had to write the code once.  I'm hoping this can cut down on the need to copy/paste to handle both floors and ceilings.

z8lfksf

### DIFF
--- a/src/MapLine.cpp
+++ b/src/MapLine.cpp
@@ -543,52 +543,49 @@ int MapLine::needsTexture()
 	if (!backSector())
 		return TEX_FRONT_MIDDLE;
 
-	// ZDoom's Plane_Align ensures a slope will hide the lower or upper texture
-	// of one side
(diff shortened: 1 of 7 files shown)


eevee:
Superseded by #296.

PR #259, opened 2015-05-19
Fix for an architecture merging oversight

eevee:
Draw some architecture like this:

    +-------+
    |       |
    |       |
    |   +---+
    |   |
    +---+

Enter lines mode, and drag the bottom wall up and to the right so it _overlaps_ the middle wall, like this:

    +-------+
    \       |
     \      |
      +-+-+-+

Result: when you let go of mouse2, SLADE freezes for several minutes.  Once it's unfrozen, if you try to undo, the result is broken — the middle vertical wall is now two-sided, facing outwards, and _sharing a side_ with the left wall!  This is totally impossible to fix without editing the map manually; if you try to fix the side reference on either line, SLADE thinks that side index is now free for the next side you create.

It turns out I trip over this _a lot_, because I'll forget that a single solid wall is actually composed of two lines, like this:

    +-----------+----------+

And I'll try to move the wall to the left or the right, and accidentally make one piece overlap the other.  Now SLADE is frozen and my map is irreversibly screwed up.

Good news: I fixed it!

----

This was all super gnarly to figure out so I hope you can appreciate this tale of diving down a rabbit hole  :)

The initial problem, the freezing, is addressed by the first commit. For pathological architecture, `traceOutline` would sometimes end up in a loop, doomed to never return to its starting point.  It gives up after ten thousand steps, which is why SLADE _only_ freezes for a few minutes.  I taught it to also never try to follow a line it's already followed once, which eliminates this problem.

Once I'd fixed that, there was another problem: the merge failed spectacularly, producing two lines that overlap and two lines with no sides whatsoever.  Ultimately this wasn't so much a bug as an oversight: `mergeArch` knows to split existing lines when a vertex moves on top of them, but it doesn't check for lines that have moved onto an existing vertex.

After fixing _that_, undoing the merge produced garbage!  And _that_ was because `MapObjectCreateDeleteUS` tries to guesstimate whether any objects were created or deleted by comparing the lengths of lists and their last elements...  but in the test case above, the splitting and merging of lines ends up replacing a line in the middle of the lines list, and the undo system
thought nothing had changed.

Now, at last, this works perfectly.

----

I also snuck in a completely irrelevant UI improvement: the "new map" dialog now defaults to the format of the first map in the wad, or the best format supported by the game otherwise.  After a few dozen trials I realized I was a little sad that SLADE wanted me to make Doom-format maps for ZDoom  :)

### DIFF
--- a/src/Main.h
+++ b/src/Main.h
@@ -114,3 +114,11 @@ enum MapTypes
 	MAP_UDMF,
 	MAP_UNKNOWN,	// Needed for maps in zip archives
 };
+
+const string MAP_TYPE_NAMES[] = {
+	"Doom",
+	"Hexen",
+	"Doom64",
+	"UDMF",
+	"Unknown",
+};
--- a/src/MapChecks.cpp
+++ b/src/MapChecks.cpp
@@ -395,11 +395,11 @@ public:
 			MapVertex* nv = map->createVertex(intersections[index].intersect_point.x, intersections[index].intersect_point.y, -1);
 
 			// Split first line
-			map->splitLine(line1->getIndex(), nv->getIndex());
+			map->splitLine(line1, nv);
 			MapLine* nl1 = map->getLine(map->nLines() - 1);
 
 			// Split second line
-			map->splitLine(line2->getIndex(), nv->getIndex());
+			map->splitLine(line2, nv);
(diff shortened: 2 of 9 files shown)


sirjuddington:
Ok looks good :) Just have one issue with a3de426 - check my comment there

PR #286, opened 2015-06-09
Double-click to edit the current selection

eevee:
Seems useful, and the notion of a double click doesn't make much sense to the keybind system.


sirjuddington:
Might be an idea to put this one behind a config (can be enabled by default)

eevee:
Sure, if you want, though I'm wary of ending up with checkbox overload — this seemed safe since currently double-click is only interpreted as two clicks, which will toggle the selection of whatever you're pointing at twice, i.e. do absolutely nothing.

sirjuddington:
I was more worried about people accidentally invoking it, I'm not too concerned about having too many options :P

eevee:
Pref added, under the "click nothing to clear selection" pref.

PR #287, opened 2015-06-09
Some slope support improvements

eevee:
Finally gave the 3.1.1 branch a go.  Fixed some places that were assuming flat flats, salvaged what I could from my other branch.  :)

I think I have an idea for making live updating of slopes work without being too invasive — it involves calculating the planes on the fly (like I did before), caching the results, and flushing the cache when any part of the sector or anything inside it changes.  All the slope mechanisms only work on the sector they're in or part of, thankfully.


sirjuddington:
Yeah that sounds like it would work ok, finding a nice way of handling live updates was one of the main things stalling 3.1.1.

PR #296, opened 2015-07-31
[WIP] Finish slope support

eevee:
In the end I had to go back to computing slopes in one big map-wide pass.  There's way too many finnicky dependencies that might affect slopes anywhere else on the map, and ZDoom's slope setup logic is very difficult to emulate correctly on a sector-by-sector basis.

It turns out this might be plenty fast enough, since it only needs to run through the map objects once rather than once _per sector_.  I popped open sunder map14 and went into 3D mode and changed some floor heights with the mouse wheel (which re-runs the slope calculations every time) and didn't notice any slowdown.  I also opened slopes.wad (the demo wad on [the wiki](http://zdoom.org/wiki/Slope), which now renders correctly!) and changed some heights and watched slopes change in realtime, which was pretty great.

This isn't quite done yet:

* `Plane_Copy` isn't fully implemented.  Vavoom slope things aren't implemented at all.  Vertex height things aren't taken into account.  (I feel like they should set the `floorz` and `ceilingz` of the vertex they're on, but that would affect what gets written into the map in the case of UDMF, even if you move the thing.)
* Various kinds of changes to map objects don't yet recompute slopes: in particular, geometry changes and ids/tags.
* Currently this code will recompute all the slopes for the whole map after every change even within a bulk edit, which hasn't noticeably slowed anything down yet but doesn't sit well with me.
* Line-based slopes for Doom-format maps won't be applied correctly.
* 3D mode doesn't correctly handle the case where two sector planes intersect on a single line — if the quad's vertices come out in the wrong order, the texture is painted in midair.  I tried fixing this with the trick of drawing concave shapes with the stencil buffer but clearly have no idea what I'm doing  :)  The cursor doesn't quite understand where to point, either.  I think this might require splitting line quads into two triangles instead, ugh.

Also I'm finding myself wishing vertices were (subtly) visible in 3D mode, so I could draw triangulated terrain and then adjust the slopes visually.


eevee:
Everyone loves screenshots.

PR #312, opened 2015-09-20
Do not reset sector mode when working in sector editing mode

Monsterovich:


sirjuddington:
What is this fixing? The intended behaviour is that you press 'S' once for sector mode (both), then can press 'S' again while in sectors mode to cycle through the sub-modes. Switching away and then back to sectors mode it supposed to reset the sub-mode, but perhaps that one can be configurable.

Monsterovich:
>What is this fixing?  

When you switch editing mode from sector mode to something else (including 3d mode), you can't keep the sub-mode because SLADE always forces it to SECTOR_BOTH.

PR #315, opened 2015-09-24
Fix some papercuts

eevee:
Some little things that I never got around to individually.  Some UI, a couple editing bugs.  Commits should be pretty independent, so feel free to cherry pick if you have any objections  :)


eevee:
Midtex comparison, current release:

slade-midtex-before

With these patches:

slade-midtex-after

In ZDoom:

screenshot_doom_20150923_162636

PR #319, opened 2015-09-25
Fix a bunch of sizing stuff with the action special panels

### LINKED ISSUE #231: Line action special window is too small by default

eevee:
This fixes #231.  Also, I hate wxWidgets.

Curiously, this same change applied to master results in an archive manager panel that refuses to go any narrower than half the width of my screen _and_ puts the two lists next to each other rather than stacked.  I'm guessing bf6d8e0a is why, but the different orientation baffles me.

From commit, for your convenience:

----

I only barely understand what is going on in most of this but this makes sizing work more better.

- `wxDataViewTreeCtrl`, used to show the list of action specials, doesn't really report a minimum size very well.  I had to manually compute the size of each of the specials' names to get the minimum width correct.

- Sometimes parts of the fancy arg controls (flag names and the like) got cut off.  Needed to explicitly recompute the top-level window's minimum size and resize it.

- `wxAuiNotebook` doesn't report a minimum size at all, which imo is not really very advanced of it.  I added a quick implementation to `STabCtrl` that's just based on the minimum sizes of the panels.

- Several panels in the action special dialog used `SetSizer()` but never set themselves a minimum size; using `SetSizerAndFit()` instead fixes this.  Probably?

### DIFF
--- a/src/ActionSpecialDialog.cpp
+++ b/src/ActionSpecialDialog.cpp
@@ -60,21 +60,31 @@ ActionSpecialTreeView::ActionSpecialTreeView(wxWindow* parent) : wxDataViewTreeC
 	// Add 'None'
 	item_none = AppendItem(root, "0: None");
 
+	// Computing the minimum width of the tree is slightly complicated, since
+	// wx doesn't expose it to us directly
+	wxClientDC dc(this);
+	dc.SetFont(GetFont());
+	wxSize textsize;
+
 	// Populate tree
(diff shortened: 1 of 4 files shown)


Monsterovich:
>I hate wxWidgets.

^^^^^^^ me too :3

sirjuddington:
Mm, not much else better though, unfortunately. wxAui is particularly terrible.

PR #322, opened 2015-09-28
Added basic level fog preview

Monsterovich:
What changed:
- Removed default global fog, added custom fog which can be configured from UDMF (fade color) or ACS scripts (Sector_SetFade).
- Fixed some sector/line update issues with color/fog in "checkVisibleFlats".
- Added new CVAR render_fog_distance (default: 1000)

fog

Fog enabled:
foge
Fog disabled:
fogd

### DIFF
--- a/src/MapRenderer3D.cpp
+++ b/src/MapRenderer3D.cpp
@@ -55,6 +55,7 @@ CVAR(Int, render_3d_things, 1, CVAR_SAVE)
 CVAR(Int, render_3d_things_style, 1, CVAR_SAVE)
 CVAR(Int, render_3d_hilight, 1, CVAR_SAVE)
 CVAR(Float, render_3d_brightness, 1, CVAR_SAVE)
+CVAR(Float, render_fog_distance, 1000, CVAR_SAVE)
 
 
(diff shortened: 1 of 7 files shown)


sirjuddington:
Hmm, I'm not sure how (G)ZDoom handles custom fog colours in relation to sector light levels, but this currently breaks any 'normal' sectors since it removes the original fog distance calculations and makes it static, which means bright sectors have the same fog distance as dark sectors, which isn't quite right (take a look at Doom2 MAP01 with this and you'll immediately notice the problem)

It should probably keep the original distance calculation if a sector doesn't have custom fog, and use the new one if it is custom.

Monsterovich:
Now it should be better. Read all changes here: https://github.com/Monsterovich/SLADE/commit/e197f5807c2eb90d4761b19fdc39036f5975ee36

and some screenshots:
Old formula:
oldformula
New formula:
newformula

Monsterovich:
Will you merge these changes or I should fix something else?

sirjuddington:
Once I've tested out the changes (looks fine from the code though)

PR #325, opened 2015-10-03
Some ActionSpecialDialog improvements

Monsterovich:
Now you don't need to re-select the item, just click  twice and slade will automatically jump to the args tab.


eevee:
This made it so clicking a new special doesn't actually change the special.

Monsterovich:
Your commit broke it.
https://www.youtube.com/watch?v=_SukzbM0xiI&feature=youtu.be

PR #326, opened 2015-10-05
Fixed missing icons

Monsterovich:


sirjuddington:
The map rename feature needs to check the game config for valid names, and check that the new name entered is not already in use. I'd use the already existing NewMapDialog (https://github.com/sirjuddington/SLADE/blob/3.1.1/src/MapEditorConfigDialog.cpp#L55) for this instead, and modify it to be able to hide the map format dropdown.

Monsterovich:
Okay. Merge another small commit please.

PR #344, opened 2015-11-17
Misc 3.1.1 fixes

eevee:
I did a lot of mapping recently and took notes on things that were broken or annoying and here you go.


eevee:
Don't worry, I seem to be really good at finding edge cases  :D

PR #347, opened 2015-11-17
Small bugfix

Monsterovich:
https://github.com/sirjuddington/SLADE/pull/325

### DIFF
--- a/src/ActionSpecialDialog.cpp
+++ b/src/ActionSpecialDialog.cpp
@@ -1212,7 +1212,17 @@ void ActionSpecialPanel::onSpecialItemActivated(wxDataViewEvent &e)
 {
 	// Jump to args tab, if there is one
 	if (panel_args)
+	{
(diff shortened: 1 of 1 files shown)


eevee:
Well that's bizarre.  It works fine here; a double-click fires both events.

Monsterovich:
No, it doesn't. http://www.youtube.com/watch?v=_SukzbM0xiI

eevee:
I assure you the same sequence of events had different results for me  :)  Which is even all the more strange given that we both seem to be using wxGTK.

Monsterovich:
>a double-click fires both events.
I hope you know that only onSpecialItemActivated has a double-click trigger. I had these lines for it, but you returned onSpecialItemChanged by renaming onSpecialItemActivated (this is done right) and moved onSpecialItemActivated. I just want to mention that onSpecialItemChanged works only when you switch your selection from one item to another, so it doesn't work when you click any item first time. (see video)

eevee:
Oh, wait, I see; I didn't realize you'd _typed_ the first special in your video.

PR #355, opened 2015-12-01
Timidity support doesn't work on linux

Monsterovich:
Works on win7-8.
Still doesn't work on XP.


Gaerzi:
Using just program->Kill(pid) instead of specifying SIGKILL doesn't work for me. I get multiple timidity processes this way when opening several MIDI entries one after the other.

Monsterovich:
Fixed.
``
#ifdef WIN32
			program->Kill(pid, wxSIGKILL, wxKILL_CHILDREN);
#else
			program->Kill(pid);
#endif
``

Gaerzi:
isPlaying() uses the same program->Exists(pid) logic. Only difference is that it checks for 0 pid. Is that what you encountered when the wrong process was killed?

Monsterovich:
>isPlaying() uses the same program->Exists(pid) logic. Only difference is that it checks for 0 pid. Is that what you encountered when the wrong process was killed?

If pid is 0, it kills slade process instead of timidity. I checked this condition in previous version.

and I also found this on wxWidgets wiki:
> Note that under Unix, for wxKILL_CHILDREN to work you should have created the process passing wxEXEC_MAKE_GROUP_LEADER.

Monsterovich:
Can someone merge this?

Monsterovich:
Ok, I did as you wish.

PR #360, opened 2015-12-14
Some fixes to the dir archive check

### LINKED ISSUE #343: Saving a directory still sometimes deletes new files
I create a new file manually, I see the "modified files" check run several times in SLADE, but nothing happens.  Later I mash Ctrl-S in SLADE out of habit and the new file is silently deleted.

It's probably a pain since the code was all designed to work with archives, which are read and written as single units...  but we should really really not ever delete files from a directory unless they were explicitly deleted within SLADE itself.

eevee:
Fixes #343.

The "only check when app is activated" bit doesn't actually work on GTK, due to [a wx bug](http://trac.wxwidgets.org/ticket/17260) (allegedly fixed, but not in a release).  The check will also run at some silly times like when launching a modal dialog.

I made `DirArchive` remember which changes were _un_checked, though, and you won't be prompted about those changes until the same files are touched again.  So running the check more than necessary should be harmless and invisible.

Also fixed a couple other nitpicks I've made about the check, and some assertion errors I bumbled into.

Other thoughts that I punted on as they're a little more invasive:

* If a file changes outside SLADE, and isn't modified or open inside SLADE, is there any reason not to just update it silently and automatically?

* There's no real handling for files that are modified _both_ inside and outside SLADE — the change dialog doesn't mention it and will silently overwrite the internal changes, and saving the directory will silently overwrite the external changes.  You're probably asking for trouble if you're doing this, granted, but it could at least stand to have a warning or something.


sirjuddington:
Yeah it's probably fine to silently update the file in SLADE if it is modified outside, but perhaps can be an option.

As for clashing changes, a warning and/or option dialog asking which change to keep is probably best, doesn't seem worth it to do anything too complicated here.

PR #365, opened 2015-12-19
Exclude __FreeBSD__ systems from outputting stack traces OnFatalExcep…

kevans91:
…tion

I don't know if this the preferred way to handle this, but this is the only roadblock to compilation and functionality on FreeBSD. It seems like a sticky situation, but other attempts to  get wxStackWalker stuff to compile+work on FreeBSD did not turn out so well.

### DIFF
--- a/src/MainApp.cpp
+++ b/src/MainApp.cpp
@@ -128,7 +128,7 @@ EXTERN_CVAR(Bool, map_show_selection_numbers)
  * getTraceString(). wxStackWalker is currently unimplemented on mac,
  * so unfortunately it has to be disabled there
  *******************************************************************/
-#ifndef __APPLE__
+#if !defined(__APPLE__) && !defined(__FreeBSD__)
 class SLADEStackTrace : public wxStackWalker
 {
 private:
@@ -213,7 +213,7 @@ public:
 
 	~SLADECrashDialog() {}
 };
-#endif//__APPLE__
+#endif//!__APPLE__ && !__FreeBSD__
 
 /*******************************************************************
  * FUNCTIONS
@@ -876,7 +876,7 @@ int MainApp::OnExit()
 
 void MainApp::OnFatalException()
 {
-#ifndef __APPLE__
+#if !defined(__APPLE__) && !defined(__FreeBSD__)
 #ifndef _DEBUG
 	SLADEStackTrace st;
 	st.WalkFromException();


Gaerzi:
So, since you're closing this, did you manage to compile on FreeBSD? How did you solve the problem?

kevans91:
Gaerzi: I did indeed (with this fix alone), but I'm verifying this against the 3.1.1 branch instead and peeking further at alternatives. backtrace(3) is available since FreeBSD 10, so I'd like to verify that there's definitely no way it can be enabled and successfully compile before shooting for this approach.

kevans91:
Apologies for the double post, but as a final note on this:

Things would have gone smoothly if I would have actually followed the [Unix Compilation notes](https://github.com/sirjuddington/SLADE/wiki/Unix-Compilation) and compiled everything from source -- however, all of the listed dependencies are available in the Ports tree and able to be installed via `pkg`, so I was attempting to do it this way. The ports tree's default configure flags for wxWidgets (`x11-toolkits/wxgtk30`) are almost spot-on, but they also include a `--disable-backtrace` by default. Notes say that it was because of a conflict with `devel/libexecinfo` that provided a backtrace mechanism before the introduction of backtrace(3) in FreeBSD 10. 

I may ultimately try to take this up with the FreeBSD ports folks. I'm testing the build of `x11-toolkits/wxgtk30` with `devel/libexecinfo` to see if there's actually any conflict left, then I intend to try and get the default `--disable-backtrace` removed if none of the folks over there have issue with it. Following that, I'd like to get SLADE into the ports tree, if for nothing else but to improve accessibility.

TL;DR Not SLADE's problem, no action required after all.

danfe:
I've fixed `x11-toolkits/wxgtk30` port to have `--enable-backtrace` by default as of [r407359](http://svnweb.freebsd.org/ports?view=revision&revision=407359), and added `games/slade` to the ports tree.

kevans91:
Heh, so much for this then, I suppose:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=205476

=)

Kyle Evans, Student Web Developer
College of Engineering, Kansas State University
kevans91@ksu.edu    (785) 532-4643    30 Seaton Hall

danfe:
Thanks for the heads-up, I've closed that PR.

Issue #421, opened 2015-12-29
Automated tests


eevee:
Testing the UI is probably a massive can of worms, but it would be nice if I could, say, change some math stuff and be reasonably sure I didn't break it.  ;)

You could even design a fuzzer for architecture merging — make some sequence of random edits and verify that the map checks all pass after each one.

I have no idea how this is usually done with C++.  #420 might help, since the guts would be exposed as a public API?

PR #445, opened 2016-02-01
Fixed bug: conversion from Mus -> Midi is not working anymore

Monsterovich:
Someone forgot to change these lines


eevee:
This is why we need #421  :)

Gaerzi:
I've already fixed it, I don't want merge conflicts.

PR #447, opened 2016-02-01
Even more papercuts

### LINKED ISSUE #446: Archive::renameEntry doesn't work properly in directory archives
I don't know how to fix this issue in the right way, so I posted this here.

sladebug2

After I tried to convert this DS to wav, Archive::renameEntry didn't rename the entry, it just created GIBS2.lmp instead, but doom sound file GIBS2 is still there, mc shows that. 
This bug happens with ALL entry operations and only in dir archives (works with pk3 and wad).

Enjoy the mess!
http://www.youtube.com/watch?v=Xu4MIu8-7Gs

eevee:
I'm running low on papercuts!  Nice.

Second commit fixes #446, too.  (There's a looming question of what should happen if you delete entry A, then recreate it.)

### DIFF
--- a/dist/res/config/ports/include/specials_zdoom.cfg
+++ b/dist/res/config/ports/include/specials_zdoom.cfg
@@ -6,10 +6,91 @@ action_specials
 {
 	clearexisting;
 
-	arg generic_speed
+	// These keywords are from Boom's generalized specials
+	arg door_speed
 	{
 		name = "Speed";
 		type = speed;
+		custom_values
+		{
+			16 = "slow - standard Doom door";
+			32 = "normal";
+			64 = "fast - fast Doom door";
+			128 = "turbo";
+		}
+	}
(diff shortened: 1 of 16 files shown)


Gaerzi:
Here's a papercut of mine, then. When using right-click->edit properties on anything, it always default to using the special tab (if there). But in many case it's not the tab I want to have. In fact generally I'll want to edit the spawn flags instead.

First, it's a bit odd that it isn't "General" that's the default tab, since it's the first tab and it's the one that's named "General"; but secondly I'd rather have the interface remember which tab was on the last time and choose that one again.

eevee:
Oh, good point.  I've found that weird for a while but never thought to look into it, and at a glance I'm not actually sure why it's happening.  Nothing explicitly sets the default tab, and the special tab isn't the first or the last added.

PR #450, opened 2016-02-07
Reorganise source files

sirjuddington:
Rearrange all the files :P Apologies if this breaks stuff contributors are working on, it very likely won't happen again.

The #include optimisation stuff cuts down compile time a bit, but not as much as I would have liked. Oh well.

I'll merge this in to master once 3.1.1 is released and the 'stable' branch is created.


eevee:
Whoa-ho, sweet.

Git is pretty good about merging renames, so hopefully this will go smoothly.

Gaerzi:
Wouldn't it have been better to merge this into 3.1.1 so that the "stuff contributors are working on" can be adapted already and not create conflicts later?

sirjuddington:
Ok recreated this one and merged it to 3.1.1 (which is kinda like 3.1.2 branch for now, I guess)

PR #470, opened 2016-03-05
OS X fixes for wxWidgets 3.1

alexey-lysiuk:
Set install prefix only for wxGTK targets (there is no need in this on OS X)
Generic way to get executable path from bundle (works in 3.0 and 3.1)
Fixed opening of recent items from non-WebView start page (not really OS X specific)


sirjuddington:
This one might be best for the `stable` branch (#468 probably should have been done there, too)

alexey-lysiuk:
OK, I was not aware of merging `stable` branch into `master`.

PR #471, opened 2016-03-05
OS X fixes for stable branch

alexey-lysiuk:
Adjusted path to Xcode prefix header
Set install prefix only for wxGTK targets
Generic way to get executable path from bundle


sirjuddington:
This is still set to master (why doesn't GitHub just let you change the target branch of a PR after it has been made?)

alexey-lysiuk:
My bad, didn't check the base branch :disappointed:

PR #472, opened 2016-03-05
Fix for non-WebView recent items

alexey-lysiuk:
Fixed opening of recent items from non-WebView start page


sirjuddington:
This should probably also be for stable

alexey-lysiuk:
The code from the `stable` branch works fine. The problem is in `master` only.

sirjuddington:
Oh yeah, it was due to the external editing stuff.

PR #479, opened 2016-03-07
Various improvements to the text editor

sirjuddington:
A bunch of improvements and additions to the text editor:

* Added code folding feature, can fold code blocks (eg. between `{` and `}`), preprocessor blocks (eg. `#region`/`#endregion`) and comment blocks.
* Added option to override the text editor font globally (rather than using what is defined in the style)
* Added a simple preview to the text editor styles preferences page
* Moved everything from the bottom TextEntryPanel toolbar to a new 'Text' entry menu (word wrap, jump to, find+replace)
* Indent guides and right margin line colours are now configurable
* Revamped the Find+Replace feature:
  * Is now a panel that pops up below the text editor, so it doesn't get in the way of the text like the old dialog did
  * 'Find Previous' button
  * Extra matching options - match start of word, regular expression (isn't 'real' regex, but scintilla's version of it), and backslash expressions
  * Can press enter while the find text is highlighted to find next (also shift+enter to find previous)
  * When activated, the find text will be populated by either the current selection or the word under the cursor, and the 'find' text box will be focused and selected
* Added 'Jump to Line #' feature
* Vastly improved calltips:
  * No longer look like something straight out of windows 95
  * (Optionally) colourise function name and parameter types
  * (Optionally) dim function parameters that are optional
  * Limit width of calltips and split to multiple lines so they are less lilely to extend off the screen (they still can, though)
  * Pressing escape while a calltip is open will close it
  * Pressing up/down while a calltip with multiple function parameter sets is open will cycle through the sets (can also click on the up/down arrows as before)
  * Corrected some issues where calltips would hang around when they shouldn't or not show up when they should
  * Added ability to define function descriptions, which will show up in calltips
  * Changed function config behaviour to override a previously defined function if another with the same name is encountered, rather than adding the new args as a set (so no more 2x identical arg sets for ACS_Execute, etc.)
* Changed the 'Jump To' option/dialog to an auto-updating dropdown that is always showing on the top toolbar
* Fixed refresh issues with multi-line brace matching in windows
* Fixed text editor stealing focus from other controls (most noticeable in text controls - the cursor would show up in the wrong spot)


Gaerzi:
Can search & replace also be made to handle backslash expressions, so we could do things like replacing ",\t" by ",\n\t" for example? Between that and some basic regexp support, I'd need to select all, copy, then paste in pnotepad a lot less.

something like this I guess

sirjuddington:
Don't see why not, shouldn't be too difficult to do

ZwipZwapZapony:
In relation to the word wrap function thingie for the text editor... Can we possibly get an option in Edit > Preferences > Text Editor to enable word wrapping by default?
Even with a 1080p monitor while running Slade maximized, I very often (every single time) find myself ticking word wrapping on when I open an SRB2 Lua script in a Wad.
It's not really annoying for me, just a minor inconvenience, but I'd still be happy to see an option for it.

PR #483, opened 2016-03-19
Papercuts 6

eevee:
Fixed some things I stumbled upon while mapping for DUMP 2.

The only not-quite-trivial change is the "enclosed void" one.  I hammered on it for a bit and couldn't come up with any pathological cases that broke catastrophically.

### DIFF
--- a/src/Archive/Formats/DirArchive.cpp
+++ b/src/Archive/Formats/DirArchive.cpp
@@ -628,7 +628,12 @@ void DirArchive::updateChangedEntries(vector<dir_entry_change_t>& changes)
 
 		// Deleted Entries
 		else if (changes[a].action == dir_entry_change_t::DELETED_FILE)
-			removeEntry(entryAtPath(changes[a].entry_path));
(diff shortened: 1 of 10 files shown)


Monsterovich:
When 3D floors? :P

eevee:
Eventually!  They're rather a lot of work.

sirjuddington:
Only one I'm concerned about is f0687a6, the setModified shouldn't be going at the top of the function(s), but still be before any change is made. As it is here a line will be set 'modified' even if you are just setting a side or vertex property. It'll also end up with setModified called twice if it gets down to MapObject::setXXXProperty, which shouldn't matter but might.

eevee:
Aha, yeah, I'll put it after the check for side properties in the line case.

I don't _think_ calling `setModified` twice should hurt anything; the second call should be a no-op unless the global backup timestamp has changed, which definitely shouldn't be happening while we're still in the middle of a `MapObject` method call.

PR #517, opened 2016-05-02
Remove unused "ancientzip" code

zenvoid:
Remove files ancientzip.cpp and ancientzip.h. The license is not specified and its methods seems to be not used within Slade anyways.


zenvoid:
Tested compilation under Linux. Since it was unused code it should not break any other OS build, I hope.

PR #526, opened 2016-05-28
Normalize overlay actor texture size

Monsterovich:
From:
lolslade
To:
lolslade2


sirjuddington:
Might be best to merge this one into stable instead of master

PR #541, opened 2016-06-11
Fixed crash when opening lump in text editor

alexey-lysiuk:
Opening ZDoom Texture Definitions lump no longer leads to a crash


sirjuddington:
Heh, oops. Perhaps I should switch to c++11 a bit sooner than planned to avoid stuff like this.

PR #547, opened 2016-06-18
Updated the Eternity config to have parity with the more recent Eternity builds.

Altazimuth:
Previously all lines that were >400 were not in the config. They are now accounted for.
On top of this ExtraData specials have their own little group, and special 307's description was changed to reflect its current name.

### DIFF
--- a/dist/res/config/ports/include/specials_eternity.cfg
+++ b/dist/res/config/ports/include/specials_eternity.cfg
@@ -3,9 +3,14 @@
 
 action_specials
 {
+	group "ExtraData"
+	{
+		special 270 = "-- ExtraData Control Line";
+		special 401 = "-- ExtraData Sector Control";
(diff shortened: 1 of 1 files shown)


Gaerzi:
Thanks, merged.

However I'm not sure how useful it is to list the parameterized types yet since AFAIK you still have to go through ExtraData to use them at the moment. Besides, perhaps their parameters should be included? (Look at the ZDoom equivalent for an example of the syntax.)

Altazimuth:
I'll definitely look into it. Not sure how soon because I'm juggling this alongside attempting to help UDMF implementation for EE (which I have implemented partially for SLADE in the main branch of my fork). Ideally I'll make a pull for that once UDMF for EE is in a publicly workable state, and with it will come the devaluing of ExtraData.

Issue #542, opened 2016-06-12
Invalid sectors after dragging geometry


m-x-d:
**SLADE Version**: 3.1.1.2
**Operating System**: Windows 7 x64

**Issue Details**:
1. Open the attached map.
2. Set grid size to 2 mu.
3. Select vertex 3.
4. Move it 2 mu to the left.
Result: broken sectors.
Also, undoing vertex move doesn't fix sectors.

![slade_drag_bug](https://cloud.githubusercontent.com/assets/7252064/15994443/8ce8f384-310d-11e6-9348-4066a5ac26c3.jpg)
[drag_bug.zip](https://github.com/sirjuddington/SLADE/files/311034/drag_bug.zip)

eevee:
I'll have a look; this is pretty up my alley.

eevee:
Undoing works for me on a stable build, fwiw.

eevee:
Think I've fixed it; explanation in #548.

PR #548, opened 2016-06-19
Fix sector drawing bug in #542

eevee:


eevee:
I didn't want all this to end up in the merge commit, so I'm adding it as a separate comment.

Here's the test map from #542, with line indices in red.  The green vertex 3 is the vertex that moved, and thus the place that `SectorBuilder` starts tracing sectors.

slade-geometry-edge-case

`SectorBuilder` starts by tracing around the outside (front sides) of lines 7, 2, and 1.  So far, so good, but now it's run out of lines.

It knows it has an outward-facing outline, so it calls `findOuterEdge` to look for the outer boundary of this sector.  It's not immediately obvious (and I've fixed this with a comment), but `findOuterEdge` works by firing a ray due east of the vertex and looking for the first line it intersects.

The green line is that ray.  It intersects _both_ lines 8 and 9, because they share a vertex that lies exactly on the ray.  "Nearest" is judged purely by the distance to the point on the ray, so those lines _tie_ for nearest — and line 8 ends up winning, because its lower index means it's encountered first.

The next part of the sector is thus line 8.  The last thing `findOuterEdge` does is check which side of that line vertex 3 is on.  It's on the back, so tracing continues from _the back of line 8_.

Now we have a problem: the sector builder will trace the _inside_ of the right triangle and stop, pairing it with the _outside_ of the left triangle, producing completely nonsense sectors.

My fix was to introduce an explicit tiebreaker: when two lines both have the nearest intercept point, then fall back to the more heavy-handed `distanceToLineFast`.  Line 9 is obviously closer overall, so the problem is fixed.

The change only affects cases with ties, so it should be pretty safe.  It's possible there are other buggy cases that even this tiebreaker can't handle, but offhand I can't think of any.  Maybe a situation involving lines that cross each other without having a vertex...  but that way lies madness, regardless.

PR #569, opened 2016-07-22
Add Palette Gradient Tool

jmickle66666666:


jmickle66666666:
This addresses issue #569

sirjuddington:
All looks good except a minor thing, looks like your text editor was set to use spaces for indentation, where the rest of the SLADE source uses tab characters.

jmickle66666666:
@sirjuddington Ah thanks, forgot to check that situation before I started. I had spotted and fixed the typos but managed to lose those fixes when messing around with Git. All sorted now.

PR #577, opened 2016-08-01
Make SLADE's map editor recognize full path textures.

Talon1024:
I've modified the resource manager code so that walls and flats in a UDMF-format map that use the full path name of an image file from a ZIP/PK3 archive as a texture name (you'll find countless examples in WolfenDoom: Blade of Agony) will be recognized and loaded by SLADE's map editor.

### DIFF
--- a/src/General/ResourceManager.cpp
+++ b/src/General/ResourceManager.cpp
@@ -272,33 +272,32 @@ void ResourceManager::addEntry(ArchiveEntry* entry)
 			return;
 
 		// Check for patch entry
-		if (type->extraProps().propertyExists("patch") || entry->isInNamespace("patches")) {
+		if (type->extraProps().propertyExists("patch") || entry->isInNamespace("patches"))
+		{
 			patches[name].add(entry);
 			/*
 			if (name.Length() > 8) patches[name.Left(8)].add(entry);
-			if (!entry->getParent()->isTreeless()) {
+			if (!entry->getParent()->isTreeless())
 				patches[path].add(entry);
-			}
 			*/
 		}
 
 		// Check for flat entry
-		if (type->getId() == "gfx_flat" || entry->isInNamespace("flats")) {
+		if (type->getId() == "gfx_flat" || entry->isInNamespace("flats"))
+		{
 			flats[name].add(entry);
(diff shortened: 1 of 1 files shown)


sirjuddington:
Haven't had a chance to try this out yet, but just a minor thing with the formatting - the SLADE source has opening braces `{` on the next line, not the same line as you have here

Talon1024:
OK, I changed the code style of my additions, and I removed some extra code (trimmed entry names) because I'm not sure that code belongs there.

Also, could this feature be added to the stable version of SLADE as well?

sirjuddington:
If you want it in stable you'll need to create a new pull request to apply to stable instead of master.

eevee:
Whoa, I didn't even know this was possible; I'd looked for mentions of it on the ZDoom wiki before, but only found vague mention that "in the future" paths will maybe be supported somehow.

Does this imply that it should be possible to select textures by path, somehow?  At least if their names are longer than 8 characters.

Talon1024:
> Does this imply that it should be possible to select textures by path, somehow? At least if their names are longer than 8 characters.

I think SLADE already uses textures with names longer than 8 characters. Load up C1Train from WolfenDoom: Blade of Agony, and you'll see a few missing textures, despite the fact that those texture names are only 8 characters long. This is because the PNG files corresponding to those textures have "basenames" longer than 8 characters.

On the other hand, I think users should have the option of whether or not to use full path texture names, except in the case where you have two textures with the same filename, but in different folders within the PK3.

Talon1024:
Heya. Is there anything else I need to do before this gets merged?

sirjuddington:
Oh, well yeah it works fine, I thought you wanted to swap it over to stable though?

Talon1024:
I've changed my mind. It looks to me like there won't be many more changes to the stable branch before the next version comes out.

sirjuddington:
The stable branch is actually still active, and there will definitely be more 3.1.1.x builds before 3.1.2 is ready (it's still a ways off).

So I think this is actually better for stable as a temporary fix for now. As it is, even with this fix, UDMF full path textures aren't really supported properly (they display, but don't work correctly in other places such as the browser). Adding full support will probably require a bit of a rework in how SLADE handles textures, which would be for 3.1.2 or 3.1.3.

PR #625, opened 2016-12-28
Eternity UDMF Support

Altazimuth:
Also has various updates. Sorry about some of the commit messages being a bit odd.


Gaerzi:
Not really fond of some of the code duplication this causes, but it'll do for now.

Altazimuth:
If you wanted, I could make it so this stuff isn't hard-coded any more and have it delegate to configs. This would remove the need for the code duplication. The only thing I'm concerned about would be performance; is there any easy way to cache that sort of thing up until the map type changes (e.g. if I went from working in EE UDMF to ZDoom in Hexen or something) that's already in SLADE?

PR #626, opened 2016-12-29
a bunch of optimization stuff

anotak:
disable archive list width update while pasting

### DIFF
--- a/src/Archive/ArchiveEntry.cpp
+++ b/src/Archive/ArchiveEntry.cpp
@@ -48,6 +48,7 @@ ArchiveEntry::ArchiveEntry(string name, uint32_t size)
 	// Initialise attributes
 	this->parent = NULL;
 	this->name = name;
+	this->upper_name = name.Upper();
 	this->size = size;
 	this->data_loaded = true;
 	this->state = 2;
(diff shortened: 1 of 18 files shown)


anotak:
er some other of my optimizations apparently got merged into this pull request? i'm sort of bad at github, sorry, but i was trying to submit those too

src/Archive/ArchiveEntry.cpp
```diff
@@ -113,11 +117,28 @@ string ArchiveEntry::getName(bool cut_ext)
 	// Sanitize name if it contains the \ character (possible in WAD).
 	string saname = Misc::lumpNameToFileName(name);
 
+	return saname.BeforeFirst('.');
```
Gaerzi: Not sure about this. It'd essentially make the dot an invalid character for lump names. I think BeforeLast would be closer to the previous behavior of turning it into a filename and cutting out the extension.

anotak: ok, you're right that wxfilename appears to grab the last dot not the first dot, fixing, but also, question: wouldn't this still mess up . in lump names?

PR #629, opened 2017-01-05
Remove hard-coding of a wealth of UDMF rendering features

Altazimuth:
This makes it so that a great deal of features that ZDoom and Eternity UDMF have are not hard-coded string checks against the namespace. Instead they are delegated to the config files. This avoids code duplication, and provides an example of what can be done in future to avoid code duplication when EE or ZDoom get their own specific UDMF features that work their way into being rendered.
A couple things still are hard-coded though.

### DIFF
--- a/dist/res/config/ports/eternity.cfg
+++ b/dist/res/config/ports/eternity.cfg
@@ -43,6 +43,16 @@ port eternity
 
 	// UDMF namespace
 	udmf_namespace = "Eternity";
+
+	udmf_slopes = true;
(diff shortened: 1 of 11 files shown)


Gaerzi:
Not sure why udmf_side_scaling exists. Is it planned for Eternity to have a common scale for all textures on a sidedef?

Altazimuth:
Yeah. Per-texture-scaling will probably result in rendering code triplication that we don't want in, same for per-texture-offsets.

PR #649, opened 2017-02-01
More SRB2 support

SteelTitanium:
Treat soc, maincfg and objctcfg as one lump type. Allow images with certain sizes to be converted to doom flat for use as a fading mask.


SteelTitanium:
To keep a conflict from likely happening down the rode, merge #648 first, to revert the changes in the `master` branch

PR #651, opened 2017-02-03
Require C++ 11 in cmake

### LINKED ISSUE #627: CMakelists.txt maybe missing definition.
**SLADE Version**: (3.1.2.alpha), current repo
**Operating System**: (Linux, Debian 8.6)

**Issue Details**:
Many similar errors when compling:
`error: â€˜nullptrâ€™ was not declared in this scope
  UDMFPropMap* map = nullptr;`

Adding this line on line 32 of src/CMakelists turns errors into warnings.
`add_definitions(-std=c++11)`

Slade does compile with many warnings, but editing CMakelists should not be needed.

matthew-mcallister:
C++11 has to be specified to compile on Debian/Unix. This works in jessie and stretch, and should work for other systems as well.

Presumably fixes #627.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -114,6 +114,14 @@ endif(APPLE)
 # enable SSE instructions for dumb library
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -D_USE_SSE -msse")
 
+# c++14 is required to compile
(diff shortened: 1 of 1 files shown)


sirjuddington:
Should probably change this to c++14, since I'm using std::make_unique in a recent commit

matthew-mcallister:
Got it. This still compiles in jessie, which is nice.

PR #671, opened 2017-02-22
Fix pre-compiled header in cmake

matthew-mcallister:
The pch generated by cotire was ignored because of sakra/cotire#94 and because cotire must be added after all target properties of the executable have been set.

The second commit isn't part of the fix but may save some trouble someday.

### DIFF
--- a/cmake/cotire.cmake
+++ b/cmake/cotire.cmake
@@ -43,7 +43,7 @@ if (NOT CMAKE_SCRIPT_MODE_FILE)
 endif()
 
 set (COTIRE_CMAKE_MODULE_FILE "${CMAKE_CURRENT_LIST_FILE}")
-set (COTIRE_CMAKE_MODULE_VERSION "1.7.8")
+set (COTIRE_CMAKE_MODULE_VERSION "1.7.9")
 
 # activate select policies
 if (POLICY CMP0025)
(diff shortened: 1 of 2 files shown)


sirjuddington:
This seems to break the macos build (doesn't look like it sets the standard to c++14 correctly there)

matthew-mcallister:
Oh well. I don't have a mac, so I just reverted the second commit.

sirjuddington:
Still broken unfortunately, must be something to do with the first commit. The error(s) I'm getting:

```
/Users/simon/dev/slade-macos-build/SLADE/src/Archive/Archive.cpp:70:19: error: no member named 'make_unique' in
      namespace 'std'
        dir_entry = std::make_unique<ArchiveEntry>();
                    ~~~~~^
/Users/simon/dev/slade-macos-build/SLADE/src/Archive/Archive.cpp:70:31: error: 'ArchiveEntry' does not refer to
      a value
        dir_entry = std::make_unique<ArchiveEntry>();
                                     ^
In file included from /Users/simon/dev/slade-macos-build/SLADE/src/Archive/Archive.cpp:33:
/Users/simon/dev/slade-macos-build/SLADE/src/General/Misc.h:22:7: note: declared here
class ArchiveEntry;
      ^
/Users/simon/dev/slade-macos-build/SLADE/src/Archive/Archive.cpp:70:45: error: expected expression
        dir_entry = std::make_unique<ArchiveEntry>();
                                                   ^
3 errors generated.

```

matthew-mcallister:
Hmm, that's strange. Without a mac, there's only so much I can do. Could you at least give the cmake version you're using?

I'm pretty sure it has to do with the ordering of the cotire commands relative to setting other compilation properties. I tried moving the include to the top of the file, as recommended by the cotire README. If that doesn't do it, this will probably have to wait until someone else fiddles with it or I just borrow someone else's mac.

sirjuddington:
Looks like that has fixed it

matthew-mcallister:
Great, thanks; I expected worse. I re-added the latter commit. Does the PR still work?

sirjuddington:
Nope that has broken it again :P

matthew-mcallister:
Amazing. Well, this has been enough of a merry-go-round, so I've rescinded the latter commit as it is non-essential.

PR #675, opened 2017-03-01
Save config file when applying preferences

matthew-mcallister:
Unfortunately, SLADE will never be free of crashes (it already crashes a bit more frequently than I'm sure everyone would like it to), which cause the user to have to redo any changes to their config since the last boot.

I don't know if you have any specific thoughts on the issue, but this seems like a pretty satisfying solution to me.


sirjuddington:
Don't see why not

PR #677, opened 2017-03-02
cmake: Use find_packages for all libraries and small improvements

matthew-mcallister:
This changes all libraries to be included in cmake using the `find_packages` command, the main benefit of which is ~that configuration, rather than compilation, will fail if needed libraries are missing~ conciseness, it seems. (This PR was supposed to have other changes, but I scrapped them.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,7 @@ if(COMMAND cmake_policy)
 	cmake_policy( SET CMP0011 NEW )
 endif(COMMAND cmake_policy)
 
+set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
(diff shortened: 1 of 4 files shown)


sirjuddington:
Well it seems to work on osx for me, after I `brew install`'d some extra packages, though not sure if that is ok - @alexey-lysiuk would you be able to try this? I'm a bit reluctant to accept this right away because of #687

matthew-mcallister:
I just updated the branch to have a more descriptive commit message and to include the upstream glew find module, since it is not included in cmake 3.0.

matthew-mcallister:
This was merged some time ago in 8d4e7f122a8d4f89247508ebf1abbff742a2082a

PR #678, opened 2017-03-02
Correctly restore sector heights after undo

### LINKED ISSUE #654: Cannot undo changing sector heights in 3D mode.
SLADE Version: (3.1.2.alpha)
Operating System: (Linux) Debian 8.7

**Issue Details**:
Undo does not work on sector floor or ceiling height changes. 

Example: Change a few textures on wall, raise floor, press undo a few times, textures on walls change back while floor is same height.

matthew-mcallister:
This fixes #654.


sirjuddington:
Have you tried this with sloped sectors? I'm not sure it would work correctly for them

matthew-mcallister:
You had me worried, but no, this doesn't break slopes. Slopes were never broken since they are re-computed after an undo. Non-sloped floors just weren't getting their planes updated.

PR #717, opened 2017-04-27
UI / Backend Split and Cleanup (part 1)

sirjuddington:
A start on some major refactoring and cleanup of code to split the UI from the back end as much as possible. I've also taken this as an opportunity to clean/modernise/update the code a bit as I go.

Ideally the end goal of this is to have the code in a state where replacing the wxWidgets UI is possible without having to rewrite much of the backend. Not that this is ever likely to happen as it'd still be a massive amount of work :P

There are a lot of changes here and it's pretty likely some things are going to be broken as a result, so I'm going to leave this open for a bit for testing.

Some notes:
  * I've introduced an updated naming convention in some of the later commits, currently for the new classes only - private member variables in classes have a trailing underscore (eg. `variable_`). For now I'm just going to use this for new classes added, though at some point I'd like to go back and change everything to use this convention for consistency.
  * The map editor context stuff should make it easier to allow having multiple maps open at once if I ever decide to implement it
  * Currently the log message types are unused. I'd like to improve the console panel to display the log as a list with icons/colouring for the different message types, but that can come later.

###  Changes
A general overview of the changes is below, for more detail see the individual commits.

* Added a new `UI` namespace for general UI-related functions, eg. invoking the splash window
* Added a new `MainEditor` namespace to group together things related to the main (resource) editor
  * The `MainWindow` is now held in the `MainEditor` namespace (no longer a singleton)
* Split the `MainApp` class into a general `App` namespace and a wxWidgets app class `SLADEWxApp`
  * All the old `theApp->` stuff is now under the `App` namespace
* The `SAction` and `SActionHandler` stuff is now more self-contained
* `SAction` definitions are now loaded from `actions.cfg` in `slade.pk3`, rather than being hard-coded
* The logging system has been reworked to use a custom logging implementation rather than wrapping `wxLogMessage`
  * Log messages can now have a type (Info/Debug/Error etc)
  * SFML log messages should now be written to the log file
  * The console now uses the global log rather than it's own internal one
* Added a new `MapEditor` namespace for grouping together things related to the map editor
  * The old `MapEditor` class is now `MapEditContext`
  * `MapTextureManager` and `MapBackupManager` are now held in the `MapEditor` namespace
* Split up the `MapEditContext` class into several smaller classes (in the `MapEditor/Edit` folder)
* Moved most non-UI-related functionality out of `MapCanvas` to `MapEditContext` and related classes
  * Drawing is now handled by a new `Renderer` class
  * Input is now handled by a new `Input` class


Gaerzi:
To prevent conflicts I've put on hold the work I was doing on pixel editing. I have replaced the temporary brush arrays and macros with an SBrush class and my idea was that they were responsible for creating their own SAction. (To avoid having to init both the brush and the action.)

PR #733, opened 2017-06-11
Minor tweaks to SIFormat.cpp

SteelTitanium:
Just found out what the third number does, that explains why despite the changes, I still could not convert 32x32 and other sizes of images to doom flats.


Gaerzi:
There's the gfx_extraconv CVAR for when you need these unusual sizes.

Issue #676, opened 2017-03-01
3D-Mode Camera always stuck pointing at either ceiling or floor and spinning rapidly.


Seachameleon:
**SLADE Version**: 3.1.1.5
**Operating System**: OSX

**Issue Details**:
I used SLADE previously for WAD compiling and construction, and it worked wonderfully for that purpose. Unable to find any good mapping tools compatible with OSX, I also began to use the map creation tool, which, aside from an at-times odd interface, worked flawlessly. 
Since then I've had my hard drive wiped for irrelevant reasons, and upon re-installing SLADE I quickly found that the camera in 3D mode of the map viewer will constantly spin in place rapidly, pointing either as far up or far down as possible. 

The only other issue to reference this I've found is #253

fadookie:
I am also getting this 100% of the time with latest stable build on macOS El Capitan.

alexey-lysiuk:
The issue is caused by Retina/HiDPI support in wxWidgets.
Unfortunately I have no vision how to fix this in a generic way, i.e. without some hackish workarounds.

perfectionismtech:
As a temporary workaround, open the app in low resolution mode (on Get Info).

fadookie:
@alexey-lysiuk @perfectionismtech Thank you, enabling low resolution mode fixed the issue!
Until there is a proper fix for this I'd suggest disabling retina support altogether by setting `NSHighResolutionCapable` to NO in `Info.plist` ... if I have time I'll make a PR.

alexey-lysiuk:
What's the point of disabling HiDPI support for everyone?
If user needs 3D view there is a workaround: set low resolution mode in bundle's properties.

fadookie:
The point would be so that new users don't have to trawl around github
trying to find said workaround when a whole edit mode is completely
unusable on most MacBook or iMacs made after 2012... if there are users on
mac who don't need this feature at the very least I think this should be
added prominently to the documentation.

sirjuddington:
Closing - should be fixed now

Feel free to reopen the issue if needed

fadookie:
On recent versions of macOS (I'm on Sequoia 15.4.1 with SLADE v3.2.5) the 3D camera is still almost impossible to control and spins wildly. The old workaround of setting low resolution mode doesn't seem to work, the checkbox for this is gone in Finder and I even tried editing the app's `Info.plist` and adding `NSHighResolutionCapable` = `NO` but this doesn't seem to have any effect.

[...]

PR #738, opened 2017-06-17
Attempt to support HiDPI in 3D map view

alexey-lysiuk:
Came up with this not quite elegant solution for #676.
Works fine in low and high resolution modes on macOS.

**I cannot test it on other platforms so any feedback will be appreciated.**


sirjuddington:
Works fine in linux (not high DPI), will test it in windows soonish

Issue #746, opened 2017-06-29
Feature Request: Command to quickly Comment/Uncomment code in the text editor.


crisisinaptica:
**SLADE Version**: 3.1.2 beta2
**Operating System**: Linux

**Issue Details**:

I would like to suggest the inclusion of a quick way to comment and uncomment lines and blocks of code.

I like how this is managed in pycharm: [https://www.jetbrains.com/help/pycharm/commenting-and-uncommenting-blocks-of-code.html](https://www.jetbrains.com/help/pycharm/commenting-and-uncommenting-blocks-of-code.html)

If the caret is on a line of code or a block of code is selected and press " Shift + Slash" the line or block of code is commented/uncommented using line comments ( // code )
In the same way if you press " Control + Shift + Slash " the line or block of code is commented/uncommented using block comments ( /* code */ )

crisisinaptica:
I have put some work on this [Line comments implemented](https://github.com/crisisinaptica/SLADE/commit/2aa3188065437225bbaa8f492564efb4c56a07e2)

For now you can comment/uncomment lines or blocks of code using simple line comments "//"
Tomorrow I will finish the block comments functionality.

I'm no programming wizard by any means and It's my first time tinkering with the SLADE sources, so probably there are a million ways of doing this better and prettier, but it's working nice I think.
Tell me what you think :)

crisisinaptica:
Closed via 21b5cc2

PR #753, opened 2017-07-05
New comment code text editor commands.

crisisinaptica:
I finally ended up working in my own request #746 

There are two new commands for the text editor:

- Line Comment
default bind = `Ctrl + /`
If there's no selection a simple line comment `//` is added to the line the caret is on. If there's a comment in the current line, it's cleared. Then the caret jumps to the next line.
If there's a block of text selected the command will act upon all the lines that have a character selected.

- Block Comment
default bind = `Ctrl + Shift + /'`
It wraps the selected text on block comments `/* text */`
If the selection begins with `/*`, and ends with `*/`, the comment will be cleared.

Let me know if there's something I must improve or change.

### DIFF
--- a/src/General/KeyBind.cpp
+++ b/src/General/KeyBind.cpp
@@ -505,6 +505,8 @@ void KeyBind::initBinds()
 	addBind("ted_jumptoline", keypress_t("G", KPM_CTRL), "Jump to Line", group);
 	addBind("ted_fold_foldall", keypress_t("[", KPM_CTRL|KPM_SHIFT), "Fold All", group);
 	addBind("ted_fold_unfoldall", keypress_t("]", KPM_CTRL|KPM_SHIFT), "Fold All", group);
+	addBind("ted_line_comment", keypress_t("/", KPM_CTRL), "Line Comment", group);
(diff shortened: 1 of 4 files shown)


sirjuddington:
Code looks fine, will test it a bit later. One thing that is probably needed is to get the comment strings (`//` etc) from the text editor's current language, and just default to the c style ones if they aren't defined in the language.

crisisinaptica:
For completeness' sake, should I assume lua-style comments are not supported in the editor at the moment?

srb2lua.txt has two different comment styles defined, but the TextLanguage getter methods for the comment strings only return the first type defined (c-style).
In addition, lua-style comments are not styled in the editor.

```
srb2lua {
	name = "SRB2 Lua Script";
	comment_begin = "/*", "--[[";
	comment_end = "*/", "]]";
	comment_line = "//", "--";
(...)
```

I suppose I could add a special case to recognize the lua-style comments, but since the lack of support goes beyond these commands, I don't know if that's the ideal solution.

Is there something I am overlooking?

sirjuddington:
Yeah only one set of line/block comment strings is currently supported per language, the extra ones are just there in the config since they technically *should* be supported, it just hasn't been implemented yet

Issue #767, opened 2017-08-05
I have my own build where I've fixed some bugs


usernameak:
Can i give it to you and you'll try to merge that? But there's a problem. It uses wxGTK3 that is supported only on Arch Linux. And for some reasons links in the WebView don't work.

The changes:

- Doom Builder style borders on the things in 3d mode (thin bounding box and arrow)
- Now sprites that are scaled through TEXTUREx are scaled in map editor
- GTK3
- Doom Builder style hilight in 3d mode
- Fixed a bug in TEXTUREx editor when pressing Enter is required to save the changes
- Buggy support of translations for sprites

sirjuddington:
It's best to fork SLADE on GitHub and create a pull request.

I already added GTK3 support in the gtk3 branch, including a cmake option to enable/disable it (if you have the gtk2 build of wx). Not sure why I didn't merge it in yet.

PR #771, opened 2017-08-08
Pull request for issue #767

usernameak:
See #767


sirjuddington:
I think this will need to be split up into smaller pull requests, since it does a lot of different things, some which are fine and some which I would want changed before accepting them.

Some quick notes:
- The DECORATE changes can wait a bit since they would need to be reimplemented anyway once the [mapinfo](https://github.com/sirjuddington/SLADE/tree/mapinfo) branch is merged
- The changes to the map editor visuals (3d mode hilight, things, etc.) should be optional
- Try to keep to the existing code style (opening curly braces on the next line, 4-width tabs for indentation, spaces after if/for/etc. statements)

PR #786, opened 2017-09-09
Map editor: add angle information about linedefs and change format of line length.

Dethernal:


Dethernal:
Also change FindItemByPosition(0) with GetMenuItemCount() because first give me assertion.

PR #848, opened 2018-02-05
Add appdata

Alexander-Wilms:
This adds an [AppData](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps) file, which makes applications look better in Gnome Software & KDE Discover.

Also, this changes the ID of the desktop file to use the same reverse domain name notation.

### DIFF
--- a/net.mancubus.SLADE.appdata.xml
+++ b/net.mancubus.SLADE.appdata.xml
@@ -1,0 +1,94 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+	<id>net.mancubus.SLADE</id>
+	<name>SLADE</name>
+	<summary>It's a Doom editor</summary>
+	<metadata_license>CC0-1.0</metadata_license>
+	<project_license>GPL-2.0</project_license>
+	<developer_name>Simon Judd</developer_name>
+	<url type="homepage">http://slade.mancubus.net/</url>
+	<url type="bugtracker">https://github.com/sirjuddington/SLADE/issues</url>
+	<description>
+		<p>
+			SLADE3 is a modern editor for Doom-engine based games and source ports. It has the ability to view, modify, and write many different game-specific formats, and even convert between some of them, or from/to other generic formats such as PNG.
+		</p>
+		<p>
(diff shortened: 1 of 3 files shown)


net.mancubus.SLADE3.appdata.xml
```diff
@@ -0,0 +1,94 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+	<id>net.mancubus.SLADE3</id>
+	<name>SLADE3</name>
+	<summary>It's a Doom editor</summary>
+	<metadata_license>CC0-1.0</metadata_license>
```
Alexander-Wilms: Is llicensing the matadata under CC0 ok? Citing the AppData spec:
"A permissive [metadata-]license ensures your data can be combined with arbitrary other data in one file, without license conflics (this means copyleft licenses like the GPL are not suitable as metadata license). Possible license identifiers include:

    FSFAP
    CC0-1.0
    CC-BY-3.0
    CC-BY-SA-3.0
    GFDL-1.3
    MIT"

sirjuddington:
Looks fine, although I'd probably just go with SLADE instead of SLADE3 for the name. Also with the id, does it have to be the website? Seems a bit strange to use that as it could potentially change during the lifetime of the application

Alexander-Wilms:
Alright, should the description also mention SLADE, or SLADE3?

Yes, the first two elements of the ID must be an actual website, so a lot projects use e.g. com.github.sirjuddington.SLADE

cf. the AppStream spec:

"The ID must follow a reverse-DNS scheme, consisting of {tld}.{vendor}.{product}, for example org.kde.Gwenview or com.hugski.ColorHug2. Ownership of {vendor}.{tld} in the domain name system guarantees uniqueness of IDs. "

https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

sirjuddington:
Hmm well ok that should be fine then.

Keep 'SLADE 3' in the description because otherwise it doesn't make much sense (I should probably rewrite the 'about' stuff at some point), but the application name should be just SLADE.

Alexander-Wilms:
Done

PR #862, opened 2018-04-06
Make txed_{export,extract} texts match their function

Talon1024:
"Export Texture" really converts a composite texture to a single patch, and "Extract Texture" exports composite textures to PNG files. I also used "Patch" and "PNG file" because these menu items were on a submenu titled "Export To".


sirjuddington:
This one makes sense but it's probably best to do it on the master branch - the SAction definitions have been moved to a config file instead of being all in one big hardcoded function

PR #882, opened 2018-06-10
fix WAV RIFF header reported size

landfillbaby:
account for internal headers in size field (assuming only 2 chunks)


landfillbaby:
don't merge yet, i overcompensated

landfillbaby:
2 chunks, each with 8-byte header, plus the 4 'WAVE' bytes = 20

PR #894, opened 2018-07-02
Add full path texture support to the map editor

Talon1024:
This adds another section to the texture browser "Textures (Full Path)" when editing a map for a game that supports full path textures, and an archive with full path textures is open.


Talon1024:
I just removed the commit that fixes short-named textures not showing up. The same fix is in #895, along with some other additions.

Issue #903, opened 2018-08-27
Slade fails to build pk3 on Linux due to error on 7z command


dwrl:
**SLADE Version**: current git
**Operating System**: Manjaro Linux, current

**Issue Details**:

Building Slade works, but it fails at the pk3. The problem seems to be with line 57 of "dist/CMakeFiles/pk3.dir/build.make", for 7z returns an error. Crippling this file and placing "slade.pk3" from an windows release into the build directory enabled me to install a working copy of Slade.
Maybe the command line arguments of 7z have changed somehow?

Output:
```[100%] Built target slade

7-Zip [64] 16.02 : Copyright (c) 1999-2016 Igor Pavlov : 2016-05-21
p7zip Version 16.02 (locale=de_DE.utf8,Utf16=on,HugeFiles=on,64 bits,4 CPUs Intel(R) Core(TM) i5-2500 CPU @ 3.30GHz (206A7),ASM,AES-NI)

Scanning the drive:
34 folders, 733 files, 5835138 bytes (5699 KiB)

Creating archive: /home/tarl/Downloads/slade/SLADE-master/slade.pk3

Items to compress: 767



System ERROR:
E_INVALIDARG
make[2]: *** [dist/CMakeFiles/pk3.dir/build.make:57: dist/CMakeFiles/pk3] Fehler 2
make[1]: *** [CMakeFiles/Makefile2:396: dist/CMakeFiles/pk3.dir/all] Fehler 2
make: *** [Makefile:130: all] Fehler 2```

Real-Gecko:
This happens due to `-mtu+` 7z flag which is unrecoginzable. I created PR to fix the issue.

PR #904, opened 2018-09-04
Update CMakeLists.txt for quick fix of #903

Real-Gecko:
Quick fix for #903, `-mtu+` option is not recognized by 7z.


Real-Gecko:
Thanks!

PR #913, opened 2018-10-09
Fix memory leak in vertex mode (gtk3 branch)

nevesnunes:
MapRenderer2D::renderVerticesImmediate() did not call glDeleteLists(), which lead to vertices being allocated but not deleted. I applied similar checks as the ones in MapRenderer2D::renderLinesImmediate(). Probably worth applying this fix to the main branches?


sirjuddington:
Yeah it's probably best to apply this to master instead of gtk3

sirjuddington:
Closing this since it was merged to master

PR #920, opened 2018-11-03
Tokenizer improvements.

crisisinaptica:
* Do not increment `state_.current_line` if a token is ended at a newline

Currently if that's the case, 'current_line' is incremented twice.
This makes the text editor jump points to misbehave in files with a
code style like this:

```
1   function void
2   function_name ()
3   {
4     ...code...
5   }
6
7   if (test)
8   {
9     ...code...
10  }
11  else
12  {
13    ...code...
14  }
```

`state_.current_line` is incremented twice at lines 1 and 11

* Remove default case in switch statement, all possible cases are dealt
with

* Use the more robust `wxFileOffset` and `size_t` types where appropriate

* Make some implicit casts explicit

### DIFF
--- a/src/Utility/Tokenizer.cpp
+++ b/src/Utility/Tokenizer.cpp
@@ -110,8 +110,8 @@ bool Tokenizer::Token::isFloat() const
 bool Tokenizer::Token::asBool() const
 {
 	return !(S_CMPNOCASE(text, "false") ||
-			S_CMPNOCASE(text, "no") ||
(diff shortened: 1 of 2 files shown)


sirjuddington:
Looks good, though I'd rather not use wx types in non-UI code (apart from wxString for now). `unsigned` should be fine to use instead (or `size_t`)

crisisinaptica:
Done.

Issue #890, opened 2018-06-24
Add BCS language support.


crisisinaptica:
https://github.com/wormt/bcc/wiki

It would be great if BCS language support could be added to the text editor in a future version of SLADE.

PR #923, opened 2018-11-05
Fix/Enhancements for text editor calltips.

crisisinaptica:
I'm going to add more commits later. We can squash them after review if they are accepted.

### DIFF
--- a/dist/res/config/languages/acs.txt
+++ b/dist/res/config/languages/acs.txt
@@ -1,5 +1,6 @@
 
-acs : cstyle {
+acs : cstyle
+{
 	name = "ACS (Hexen)";
 	case_sensitive = false;
 	keyword_link = "http://zdoom.org/wiki/Special:Search/%s";
@@ -7,11 +8,13 @@ acs : cstyle {
 	function_link = "http://zdoom.org/wiki/%s";
 	blocks = "Script", "function";
 
-	keywords = {
+	keywords =
+	{
 		open, restart, script, special, str, suspend, terminate, world, 
 	}
 
-	constants = {
+	constants =
+	{
 		// General
 		TRUE, FALSE, ON, OFF, YES, NO,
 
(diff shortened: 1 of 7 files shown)


crisisinaptica:
@sirjuddington I  would like propose some changes... Let's say I define the next function in `acs.txt`

```
int IsMultiplayer
{
description = "Checks if the game is currently a multiplayer game.";
deprecated = "";
deprecated_f = "GameType";
}
```

1.  `deprecated_f` is the function that supersedes this function. 
If both parameters are defined the deprecation message in the tooltip should be something like `(Deprecated in v%s, use %s)`

2. If a function does not have defined args, use `void` for the tooltip.

The tooltip for the above function should then display:

`(Deprecated, use GameType) int IsMultiplayer (void)`

or

```
(Deprecated, use GameType)
int IsMultiplayer (void)
```

crisisinaptica:
Actually, `deprecated` is not even an argument of `TextLanguage::addFunction`

sirjuddington:
The `deprecated` value is just parsed from ZScript at the moment (eg. `deprecated("2.3")`). I agree it'd be nice to have something to suggest a replacement, which could be defined in the SLADE text language definition for ZScript. It would just have to add to the existing function definition from the parsed ZScript instead of overriding it completely.

crisisinaptica:
Ahhh I see, I haven't yet looked into zscript.
I'm going to use bcs for a zandronum mod... So I'm going to start working in #890 soon.

Before that I would like to work a little more in the text editor current features, in part to fix some things, in part to familiarize myself with the code.

I would like to add this `deprecated_f` parameter to the ACS function definitions and update `acs.txt` so all function definitions have their proper description, return value, etc...

crisisinaptica:
I think I'm done for now, I have another small patch that depends on a feature only present in the current development version of wxWidgets, so it will have to wait.

I'll would like to make a full update of the language definitions at some time in the future, to set all the function descriptions, etc...

zscript
acs

sirjuddington:
Looks good, only thing I would change is to remove or reduce the vertical spacing between the deprecated message and the function

crisisinaptica:
Done :+1:

PR #942, opened 2018-11-12
Support all defined comment styles for a language in the text editor.

crisisinaptica:
· The lexer now highlights all defined comment styles.
· The `TextLanguage` class now has a `preferred_comments_` variable used to indicate `TextEditorCtrl` functions which comment style should be use.
· `preferred_comments_` defaults to `0`, the first comments specified in the language definition.
· Added a new 'Cycle Comments' action (Default bind is `Alt + /`) to cycle said variable through all supported comment styles.

Currently we assume the number of comment styles for line and block comments is always the same.


crisisinaptica:
Ah! I forgot to set defaults after changing them to vectors...

Issue #929, opened 2018-11-08
Map editor 2d view misbehaves after saving map.


crisisinaptica:
**SLADE Version**: Master Branch after 6029370232c8fd22f624e51c17c71b3a6eec06e2
**Operating System**: Linux

**Issue Details**:

Upon saving a map via `Save map changes` or `ctrl + s` several things have happened to me:

* The 2d Map View stops rendering textured flats, independently of current settings and instead it renders either:
** A wireframe view.
** Sectors in the 2d Map show a repeating pattern of thing icons (usually the direction arrow or just a circle) instead of the selected flat render style.

At different levels of zoom, some sectors change from rendering gibberish to textures.
Cycling the flat types and sector modes appears to fix all this.

* Things stop rendering textures, and show empty boxes instead.

![madness](https://user-images.githubusercontent.com/21117400/48226544-5713fb80-e3a0-11e8-8efa-4a09c805446f.jpg)

Issue #918, opened 2018-10-28
SLADE has a tendency to crash when saving while map editor is open


OrdinaryMagician:
**SLADE Version**: Since commit 6029370
**Operating System**: Linux

**Issue Details**:
Saving while the map editor is open has a high chance of causing SLADE to crash. The crash dialog is completely blank (doesn't draw), so I can't give a trace.

crisisinaptica:
@OrdinaryMagician Do you mean saving the archive that the map is in, or saving the map itself?
Can you reproduce #929?

OrdinaryMagician:
Any kind of saving while the map editor is open.

I also can see that bug if it does NOT crash.

PR #947, opened 2018-11-14
Fix for issue #929, and maybe #918

crisisinaptica:
forceRefreshRenderer() needs to update the renderer when it's called.


sirjuddington:
Could you change this one to be merged into stable instead? Master is now for 3.1.3 (or whatever will come after 3.1.2) with the code cleanup stuff etc, so anything for 3.1.2 should be put into stable from now on.

Talon1024:
The changes in this commit make it extremely difficult to select stuff in the map editor 3D mode.

crisisinaptica:
Hmm, It was making the 2d mode scroll freak a little too.
Forcing the renderer to update just after saving the map seems to work OK.

PR #952, opened 2018-11-15
Fix inability to compile SLADE

Talon1024:
Fix CMake versions without policy CMP0072 being unable to compile SLADE


crisisinaptica:
After this commit, Cmake (2.12.3) starts having tantrums again over policy 0072.
I think `if (POLICY 0072)` should be changed to `if (POLICY CMP0072)` .

PR #957, opened 2018-11-17
 Fix corrupted display of application icon in Linux 

crisisinaptica:
This has been a minor annoyance for some time.

The application icon was badly displayed when alt tabbing, in the about dialog, in the window header...

[There are dozens of us!](https://forum.zdoom.org/viewtopic.php?f=232&t=24955&hilit=slade&start=2595#p1027680)

Also, the `slade.ico` file manager preview in my system also displays badly.

sladeico

This is weird, other `ico` files display normal.
Packing a new `slade.ico` using `icotool` also results in a file that displays as it should (in the file manager and when used by SLADE)

Using `icotool` to list the contents of `slade.ico` shows this:
```
$ icotool -l slade.ico
slade.ico: x_pels_per_meter field in bitmap should be zero
slade.ico: y_pels_per_meter field in bitmap should be zero
--icon --index=1 --width=128 --height=128 --bit-depth=32 --palette-size=0
slade.ico: x_pels_per_meter field in bitmap should be zero
slade.ico: y_pels_per_meter field in bitmap should be zero
--icon --index=2 --width=16 --height=16 --bit-depth=32 --palette-size=0
slade.ico: x_pels_per_meter field in bitmap should be zero
slade.ico: y_pels_per_meter field in bitmap should be zero
--icon --index=3 --width=24 --height=24 --bit-depth=32 --palette-size=0
--icon --index=4 --width=256 --height=256 --bit-depth=32 --palette-size=0
slade.ico: x_pels_per_meter field in bitmap should be zero
slade.ico: y_pels_per_meter field in bitmap should be zero
--icon --index=5 --width=32 --height=32 --bit-depth=32 --palette-size=0
slade.ico: x_pels_per_meter field in bitmap should be zero
slade.ico: y_pels_per_meter field in bitmap should be zero
--icon --index=6 --width=40 --height=40 --bit-depth=32 --palette-size=0
slade.ico: x_pels_per_meter field in bitmap should be zero
slade.ico: y_pels_per_meter field in bitmap should be zero
--icon --index=7 --width=48 --height=48 --bit-depth=32 --palette-size=0
slade.ico: x_pels_per_meter field in bitmap should be zero
slade.ico: y_pels_per_meter field in bitmap should be zero
--icon --index=8 --width=64 --height=64 --bit-depth=32 --palette-size=0
```
I have been investigating these `field in bitmap should be zero`.
Apparently icons stored as `bmp` inside an `ico` file should have some byte offsets that are commonly used, zeroed instead (reserved use).

So technically `slade.ico` is malformed.

It's said in [this wiki page](http://fileformats.wikia.com/wiki/Icon) (check note 1) that some windows tools go against the `ico` specification and write data at those offsets, but it's okay because Windows ignores these, apparently Linux does not.

It's also specified that the 256px icon should be stored as a raw `png` file fo save space (its a `bmp` in slade.ico)

Ok, so the cool solution after learning about the amazing world of icons in windows should have been to regenerate the `slade.ico`, following specs.
I did, and didn't work when mixing `bmp` and `png` icons.

I cannot create an ico file that works in Linux and I cannot test in Windows, so I think this is the best I can do.

**TLDR** :man_facepalming: : Use the png logo instead of the ico in  Linux.

**PD**: _F***_ `ico` files.


crisisinaptica:
@sirjuddington Would it be best to have `Icons::getAppIcon` instead of `App::iconFile()`?

sirjuddington:
Probably yeah, though it'd *really* be best for the icon to just work correctly :P

No idea why it wouldn't, maybe try making it all png so it isn't mixed? Always been surprised at the complete lack of any good ico-making tools. I think I used some online thing to make the current slade ico file, which I guess was only ever tested in windows :P

crisisinaptica:
I'll give it another go soon.
I've made a full set of icons (PNGs from 16px to 1024px + vector) to update the one that get installed in Linux.

Taking advantage of this I've also created a `icns` that has more sizes that the current one.
But since I do not have a mac, I cannot test it.

The lack of good ico tools (or discrepant implementations) it's because apparently its a format without official specs, it's undocumented.

Also, it's not consistent between windows versions.

[This is the best I've found... ](https://books.google.es/books?id=eTsXKHqU158C&pg=PA131&lpg=PA131&dq=undocumented+ico+format&source=bl&ots=uT1n4lxORI&sig=wkkcWaqWcUePMLDTZdIzeemBM3Y&hl=es&sa=X&ved=2ahUKEwjPzvOm8t7eAhWpI8AKHR4yBVwQ6AEwAXoECAEQAQ#v=onepage&q=undocumented%20ico%20format&f=false)

It's really a windows thing, not a portable format.

Leaving aside that we manage to create a working ico file, I think it's for the best to have the file name string in one place only.

crisisinaptica:
Ok, One would think that having a program called `icotool` was enough.

Bless you imagemagic:
```
$ convert 11_slade_1024.png -define icon:auto-resize=256,64,48,32,24,16 slade.ico
$ identify slade.ico                                                                        
slade.ico[0] PNG 256x256 256x256+0+0 8-bit sRGB 59301B 0.000u 0:00.000
slade.ico[1] ICO 64x64 64x64+0+0 8-bit sRGB 59301B 0.000u 0:00.000
slade.ico[2] ICO 48x48 48x48+0+0 8-bit sRGB 59301B 0.000u 0:00.000
slade.ico[3] ICO 32x32 32x32+0+0 8-bit sRGB 59301B 0.000u 0:00.000
slade.ico[4] ICO 24x24 24x24+0+0 8-bit sRGB 59301B 0.000u 0:00.000
slade.ico[5] ICO 16x16 16x16+0+0 8-bit sRGB 59301B 0.000u 0:00.000
```

SteelTitanium:
Would like to say nice to finally see that issue fixed, it's been happening for as long I could remember.

PR #963, opened 2018-11-20
Fix UV coordinates and scale of scaled textures

Talon1024:
Fix UV coordinates and scale for quads that use scaled textures.
Set sx and sy to the scale specified in the texture definition.
Add world_panning field to GLTexture so that the "World panning" flag is passed to the renderer, and set it in MapTextureManager.getTexture/getFlat.
Use GLTexture worldPanning to determine how UVs of quads should be manipulated.
Don't scale UVs in setupQuadTexCoords

I tested my changes on [Final NeoDoom](https://forum.zdoom.org/viewtopic.php?f=19&t=49090&hilit=NeoDoom#p848714) MAP06, which has several road signs made from midtextures.

### DIFF
--- a/src/MapEditor/MapTextureManager.cpp
+++ b/src/MapEditor/MapTextureManager.cpp
@@ -174,6 +174,7 @@ GLTexture* MapTextureManager::getTexture(string name, bool mixed)
 						h = image.getHeight();
 						sw = imgref.getWidth();
 						sh = imgref.getHeight();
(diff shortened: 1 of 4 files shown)


Talon1024:
I'm still working on this, so it isn't ready to be merged yet.

Talon1024:
From what I can see, I may be done with this, since what I can see in the 3D view is now mostly consistent with what I can see ingame. However, it would be nice if someone else could also test these changes.

PR #964, opened 2018-11-23
GTK3 Support for Linux build

sirjuddington:
It's probably about time this gets merged in and GTK3 is made the official default. GTK2 should still work fine by specifying `-DWX_GTK3=OFF` when running cmake.

There are still a few minor oddities with GTK3, but it all works for the most part


crisisinaptica:
> a few minor oddities with GTK3
Which ones?

sirjuddington:
> > a few minor oddities with GTK3
> > Which ones?

See the extra commits :P Only other thing is that the splash window doesn't show up but I couldn't figure out how to fix that yet.

PR #965, opened 2018-11-24
Fix exporting with multiple textureX entries

Talon1024:
Check to see whether the TextureXPanel is the current page in the notebook.
Also, fix a lot of assertions from the SLADE debug build by using the correct type in PatchTableListView::getItemText.


sirjuddington:
This change broke TEXTUREx copy+paste from one archive to another, so I've reverted it for now

PR #966, opened 2018-11-24
WIP: 3d floor support in the map editor 3d mode.

crisisinaptica:
@sirjuddington I'm putting this up here to better asses what has to be done.
It's far from ready. The more feedback everyone can provide about this the better.

### DIFF
--- a/dist/res/config/ports/include/specials_zdoom.cfg
+++ b/dist/res/config/ports/include/specials_zdoom.cfg
@@ -990,7 +990,7 @@ action_specials
 				custom_flags {
 					1 = "disable lighting effects";
 					2 = "restrict lighting to inside of floor";
-					4 = "pseudo-fog";
+					4 = "pseudo-fog (GZDoom only)";
 					8 = "flat floor at ceiling height";
 					16 = "use upper texture";
 					32 = "use lower texture";
--- a/src/General/KeyBind.cpp
+++ b/src/General/KeyBind.cpp
@@ -647,6 +647,7 @@ void KeyBind::initBinds()
 	addBind("me3d_generic_up", Keypress("mwheelup", KPM_CTRL | KPM_SHIFT), "Raise target 1", group);
 	addBind("me3d_generic_down8", Keypress("mwheeldown", KPM_CTRL), "Lower target 8", group);
 	addBind("me3d_generic_down", Keypress("mwheeldown", KPM_CTRL | KPM_SHIFT), "Lower target 1", group);
+	addBind("me3d_delete_texture", Keypress("delete"), "Delete texture", group);
 
 	// Map Editor 3D Camera (me3d_camera*)
 	group = "Map Editor 3D Mode Camera";
--- a/src/MainEditor/Conversions.cpp
+++ b/src/MainEditor/Conversions.cpp
@@ -401,12 +401,9 @@ bool Conversions::wavToDoomSnd(MemChunk& in, MemChunk& out)
 	// Warn
 	if (wavbps > 1 || wavfmt != WAV_PCM || fmtchunk.channels == 2)
 	{
-		if (!(wxMessageBox(
-				  S_FMT(
-					  "Warning: conversion will result in loss of metadata and audio quality. Do you wish to proceed?"),
-				  "Conversion warning",
-				  wxOK | wxCANCEL)
-			  == wxOK))
+		if (wxMessageBox(
+				S_FMT("Warning: conversion will result in loss of metadata and audio quality. Do you wish to proceed?"),
+				"Conversion warning", wxOK | wxCANCEL) != wxOK)
 		{
 			Global::error = "Conversion aborted by user";
 			return false;
@@ -622,6 +619,7 @@ bool Conversions::vocToWav(MemChunk& in, MemChunk& out)
 			datasize += blocksize - 12;
 			break;
 		}
+		// TODO: Default case unhandled
 		i += blocksize;
 	}
 	wdhdr.size = datasize;
@@ -1242,6 +1240,7 @@ bool Conversions::auSndToWav(MemChunk& in, MemChunk& out)
 				samples[i * 4 + 1] = swapval;
 				break;
 			}
+			// TODO: Default case unhandled.
 		}
 	}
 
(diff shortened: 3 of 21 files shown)


eevee:
Sorry for leaving this hanging for so long!  I haven't gotten back to Doom stuff for a while.

Here's my own TODO list, though I can't guarantee it's entirely up to date:

- Somewhat critical problem: a sector's 3D floors are ordered by height, but the selection in 3D mode is tracked by an index into that list of floors, which means the selection is wrong if you use mousewheel to move a selected 3D floor above/below another one

    Maybe I should get rid of the duplicated `extra_floor_t` struct and instead give each sector just a list of the control sectors that affect it, then identify by those instead of by "extra floor index"?  It _is_ possible to have the same control sector mapped to two 3d floors, though...

    Really wish I had faux flat/wall objects.  Really, really wish I could just pass around sector pointers and bail if they're invalid.

- Can't highlight or select an inner flat of a 3D floor

- Mousewheel to change a _regular_ ceiling height is super slow in my dump3 map; I think the full mapspecials recalculation thing is too expensive?

    The _right thing_ here is to make map specials update incrementally in real time (rather than the current approach of constantly re-scanning the entire map), but that's much more complicated and requires a whole dependency tree.  Or, at least, incremental _slopes_ would require that; incremental 3D floors would be much simpler, since there's only one way to make a 3D floor.  Might be a reasonable way to get started on incremental specials.

- There was some kind of problem I had that I can no longer remember, to do with modifying 3D floors in real-time, argh

- Currently, most of the properties are taken from the wrong place — e.g., the floor/ceiling textures should use the _control sector_'s offsets and rotations and whatnot.  For the most part, the only property of the target sector that's actually used is its geometry.

- Don't draw wall textures on lines that have the same 3D floor on both sides

- I keep abruptly popping out of 3D mode and I'm not sure why, or if it's even my fault?  It stopped for a while, then started again when I added lighting code.

- Translucent flats need to be sorted by distance in order to draw correctly

- With `flats_use_vbo` off, slopes aren't rendered.  Oops!  Fix seems straightforward, in `renderFlat`.  (Not actually related to 3D floors but this is where I noticed it.)

- Support palette blends

- Make sure transferring the sector color is correct?

- Handle light effects

    - Flats need to propagate downwards to next floor
    - Walls: oof, quads might need splitting (consider interaction with slopes too)

- Draw walls on two-sided lines

    - Support upper/lower flag as well

- Sector info overlay still needs work

    - Says "ceiling texture", "ceiling height", etc when looking at a floor; is that right?
    - Flat lighting is wrong (I think mapspecials should intervene here)
    - Doesn't know about "flat plane at ceiling height"
    - Should probably show floor type?
    - Haven't touched walls at all and not sure how that should even work

- Make all the 3D mode keybindings do the right thing when highlighting a 3D floor

    - changing wall texture, offset, scaling
    - changing flat texture, panning, scaling
    - resetting
    - changing flat height, wall height (should also understand "flat plane at ceiling height" flag)
    - changing linked lighting (flags relevant here too)
    - changing unlinked lighting
    - auto-align (!!!)
    - floodfill selection

- Did I break the error rendering of a missing flat?

- Support Vavoom floors, too

- STILL some bizarre rendering errors
    - First time in 3D mode, the inside of a floor doesn't render; second time, it does??  what?
    - Very rarely, the VBOs break when switching back to 2D mode
    - Some sectors (_without_ 3D floors!) don't have textures
    - Modifying sectors in 3D mode can corrupt the polygons in very strange ways, presumably VBOs getting crossed
    - 3D floors are now flickering between light levels; possibly trying to draw inside and outside flats simultaneously and they're z-fighting?
    - When moving the mouse directly from one 3D floor to another, the highlight briefly flickers to the floor of the actual sector rather than the 3D floor
    - 2D mode doesn't always correctly update VBOs when deleting a sector (or possibly while creating a new one, etc)

- Code itself still has a bunch of TODOs

crisisinaptica:
@eevee Super helpful! Many thanks. This is going to give me a lot to think about while reading the code.

sirjuddington:
There are a few things I was going to do in the refactor/cleanup that may affect (and/or help) this, particularly to do with how sector floors/ceilings are stored and handled. If you look at the std_string_313_refactor branch, I've added a `Surface` struct to `MapSector` which represents a floor or ceiling plane. I'll likely be doing something like this in master, in which case extra flats for 3d floors can be stored as extra `Surface`s in each `MapSector`

Ideally I'd like to completely rewrite the 2d/3d rendering code to use modern opengl and have 3d floors in mind from the ground up, but that would be a lot of work :P Though if I ever want to properly support previewing some more advanced GZDoom features in 3d mode i'd have to do that anyway :/

Gaerzi:
>- Sector info overlay still needs work
>    - Says "ceiling texture", "ceiling height", etc when looking at a floor; is that right?
For a non-Vavoom 3D floor, that makes sense, since the top of the floor is the control sector's ceiling. The terms "upper plane" and "lower plane" could be used instead but you'd have to keep the Vavoom type in mind because for them the upper plane of the 3D floor is the control sector's _floor_. (The control sector in a Vavoom floor having a negative height, with the floor above the ceiling.)

crisisinaptica:
> There are a few things I was going to do in the refactor/cleanup that may affect (and/or help) this, particularly to do with how sector floors/ceilings are stored and handled. If you look at the std_string_313_refactor branch, I've added a `Surface` struct to `MapSector` which represents a floor or ceiling plane. I'll likely be doing something like this in master, in which case extra flats for 3d floors can be stored as extra `Surface`s in each `MapSector`
> 
> Ideally I'd like to completely rewrite the 2d/3d rendering code to use modern opengl and have 3d floors in mind from the ground up, but that would be a lot of work :P Though if I ever want to properly support previewing some more advanced GZDoom features in 3d mode i'd probably _have_ to do that anyway :/

A complete rewrite with all the GZDoom bells and whistles would be the coolest thing, but since it's probably something that's not going to happen any time soon... Would It be OK (the second coolest thing) if I bring to this branch some of the changes you have done in 'std_string_313_refactor'? From that spot we could rework some of it and finish it. 

As it stands now, far from finished and unpolished, it's OK for my needs with a couple more adjustments, but leaving this to rot being so close to something release worthy would be a shame.

sirjuddington:
Hmm, I'd say wait until I add it in master, to avoid a merge conflict mess :P I am currently in the process of working on the MapEditor source files cleanup - I'll do the MapSector/Surface stuff along with it so it shouldn't be too much longer to wait.

OrdinaryMagician:
So when's this going to get picked up again?

axredneck:
@OrdinaryMagician after refactoring of Slade.

sirjuddington:
I've finally adapted this one to the latest codebase in #1358, so I'll close this PR to be continued there

PR #967, opened 2018-11-26
Fix build

crisisinaptica:
```
In function ‘bool Conversions::spkSndToWav(MemChunk&, MemChunk&, bool)’:
Conversions.cpp:1091:34: error: expression in new-declarator must have integral or enumeration type
uint8_t* nsamples = new uint8_t[numsamples * FACTOR];
```


Gaerzi:
Why was FACTOR turned into a double in the first place? It's only used as an int.

PR #991, opened 2018-12-17
Fix compilation ifndef WIN32

evhskurk:
The loop change in src/Dialogs/Preferences/BaseResourceArchivesPanel.cpp forgot the non-win32 lines, this lets me build again.
Also removed an unnecessary #else, no functional change.


Gaerzi:
Why are you changing the [path separator](https://en.wikipedia.org/wiki/PATH_%28variable%29) on non-Windows platforms? Using a colon is traditional on Unix system; whereas on Windows systems the semi-colon was chosen instead since the colon is used to identify drive letters. Your change is going to break existing configs by forcing Linux installs to follow Windows conventions.

evhskurk:
Apparently I am blind, mistook ; for :. My apologies.
The remaining changes should still be correct.

PR #1054, opened 2019-06-28
Fix misaligned next/prev image buttons for multi-image graphics entries

mykola-ambar:
Viewing multi-image graphic entries (e.g. .ART files in DUKE3D.GRP) makes the buttons move around unexpectedly in the bottom panel of the entry preview screen, sometimes putting the next/prev image buttons out of reach. This PR fixes this problem.


sirjuddington:
Hmm, is there any way to get this to work without the next/prev image buttons permanently visible? Multi-image formats are pretty rare so 99% of the time the buttons are going to be useless. If not it might be better to at least have them disabled if it isn't a multi-image format.

mykola-ambar:
They are being hid already, https://github.com/sirjuddington/SLADE/pull/1054/files#diff-e7e3d41e125db4a992f2f6105ff41f67R149-R151

PR #1079, opened 2019-10-07
Fix: Reordered release info conform standard

Eonfge:
Was working on updating the Flatpak release, and I found that the release numbers were in the wrong order. See https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html for more info.


sirjuddington:
Ah ok, I wasn't really sure what the order was meant to be heh. You'll need to retarget this to the stable branch instead of master, since that is what the release tags point to (and the flatpak should be using)

PR #1101, opened 2020-01-05
Add GL sector specials to ZDoom configuration

Talon1024:
This allows GLSector_Skybox and GLSector_NoSkyDraw to show up in the editor as valid special sector types.


Gaerzi:
I actually thought those were in already...

PR #1120, opened 2020-03-18
Rewrite MathStuff::angle2DRad to be more efficient

Talon1024:
This code is much shorter, possibly easier to understand, and possibly more efficient than the previous implementation.


Talon1024:
Testing a Javascript implementation resulted in a slight performance gain in some cases, but a performance hit in other cases. My implementation seems to be a constant-time implementation, but the original implementation has variable performance.

PR #1141, opened 2020-05-21
UI Refresh

sirjuddington:
Various updates to the UI, including a new default icon set built from SVGs so that it looks nice at all sizes.

Other improvements/updates include:

- Toolbars have received a bunch of updates:
  - Can now be laid out vertically
  - Groups can now be added to the opposite end of the toolbar
  - No longer expand to multiple rows (was causing too many weird layout issues)
  - When groups don't fit, they are instead added to a menu under an extra 'overflow' button at the end of the toolbar
  - Buttons now correctly darken when clicked
  - Various other minor things
- Creating a new archive is now done via a single 'New Archive' action, which pops up a dialog to select the archive type. Now also allows Quake PAK archives to be created
- Moved a bunch of toolbar buttons from the main toolbar to down next to the entry list. This should hopefully make it more clear that they affect the selected entries in some way
- Entry action tool buttons are now disabled if no entry is selected
- The group box around the current EntryPanel now shows the entry name in it's label instead of 'Entry Contents'
- The EntryPanel-surrounding group box is now hidden if the entry is opened in a standalone tab, which saves a bit of screen real-estate
- Entry list filter controls can now be toggled on/off
- Entry list icon size and padding is now configurable
- Added an option to condense the translation dialog a bit vertically at the cost of a smaller image preview frame
- Rearranged toolbars for palette, gfx and texturex/pnames editors
- Bookmarks improvements
  - Bookmark action now toggles the bookmark on the entry
  - Bookmarked entries are shown with bold text in the entry list
  - Added a bookmarks toolbar button with a dropdown menu to select bookmarks
  - Added keyboard shortcut for the bookmark action (Ctrl+B)


kevans91:
> For whatever reason the mouse enter/leave events are complerely unreliable in wxGTK, so we have to constantly check if they need updating otherwise they get stuck in the wrong mouseover state

Do you happen to recall any specifics here? I removed the idle timer update locally because that's a super intense operation to be running many times per second per button. On FreeBSD with wx 3.0, gtk3, I can't immediately see any hint of an issue.

sirjuddington:
Can't say I remember much specifically except pretty much what it says - that the idle event was the only way I found to reliably have buttons show the correct mouseover state. Don't remember specifically what was wrong using just mouse enter/exit events except that they didn't always seem to fire when expected

PR #1180, opened 2020-06-29
doc: join COMPILE.md's external requirement lists

jengelh:
The requirements are mostly the same; it is understood that actual
package names can vary between platforms.


sirjuddington:
I think I'd still like to keep them separate, even if they are pretty much the same

PR #1208, opened 2020-09-13
Fixed aarch64 building

### LINKED ISSUE #1206: Won't compile on aarch64
**SLADE Version**: all
**Operating System**: Anything aarch64

**Issue Details**:

This would _almost_ work on arm/aarch64/etc if not for the `dumb` library and the baked-in `cmake` C flag `-msse`.

As it is, it's impossible to build this on non-pentium CPUs because of the `sse` requirement.

dleslie:
- Removed baked-in -msse call
- Added explicit casts for some compiler warnings

Fixes #1206 

And now it works fine on my Pinebook Pro and my RPi4.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -173,9 +173,6 @@ if(APPLE)
 	set_source_files_properties(${OSX_ICON} PROPERTIES MACOSX_PACKAGE_LOCATION Resources)
 endif(APPLE)
 
-# Enable SSE instructions for dumb library
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -D_USE_SSE -msse")
-
 # Enable debug symbols for glib (so gdb debugging works properly with strings etc.)
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -D_GLIBCXX_DEBUG")
 
--- a/src/Utility/Colour.h
+++ b/src/Utility/Colour.h
@@ -14,11 +14,7 @@ struct ColRGBA
 	// Constructors
 	ColRGBA() = default;
 	ColRGBA(uint8_t r, uint8_t g, uint8_t b, uint8_t a = 255, char blend = -1, short index = -1) :
-		r{ r },
-		g{ g },
-		b{ b },
-		a{ a },
-		index{ index }
+		r{ r }, g{ g }, b{ b }, a{ a }, index{ index }
 	{
 	}
 	ColRGBA(const ColRGBA& c) : r{ c.r }, g{ c.g }, b{ c.b }, a{ c.a }, index{ c.index } {}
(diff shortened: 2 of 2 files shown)


Cacodemon345:
Can that removed x86 SSE flag for dumb library be added back in with proper checks for x86 architecture?

dleslie:
@Cacodemon345 Yep!
But it might be clunky, I think CMake needs this assistance lib to do it "properly":
http://gitorious.org/vc/vc/blobs/master/cmake/OptimizeForArchitecture.cmake

Would you be happy with just enabling it for x86_64?

Cacodemon345:
It is easy to guard the SSE flag behind a simple check of CMAKE_SYSTEM_PROCESSOR. Check if the CPU is x86 and then switch on the flag.

Of course, enabling it just for x86_64 wouldn't be of any benefit, since SLADE still provides 32-bit versions.

sirjuddington:
@dleslie @Cacodemon345 is this one still relevant? I had left it because it seemed like it still may have needed changes judging by the discussion

Cacodemon345:
The only issue here is that this PR disables SSE instructions for libdumb unconditionally on all platforms.

Blzut3:
@sirjuddington Now that Ubuntu 18.04 has reached the end of standard support meaning i386 Ubuntu is no longer supported.  In its place with the increase in interest in running ARM workstations I'm interested in providing aarch64 binaries for Slade.  Thus I'm interested in seeing this merged.

Adding a check against CMAKE_SYSTEM_PROCESSOR could work, although since there's often not a lot of care to set this value correctly (i.e. building 32-bit binaries on 64-bit hosts), it's probably better to just do a check_c_compiler_flag call and let the compiler tell you if sse is supported on the target.  Pretty marginal difference though.

@Cacodemon345 here's my suggested solution:
```
include(CheckCCompilerFlag)
check_c_compiler_flag(-msse HAVE_SSE)
if(HAVE_SSE)
       add_compile_options(-msse)
       add_definitions(-D_USE_SSE)
endif()
```

Blzut3:
After thinking about the issue for awhile I realized there was a clever work around to compile Slade for ARM without modifying the source code:  Wrap GCC and remove the flags.
```bash
#!/bin/bash

declare -a Args=("$@")
for Arg in "${!Args[@]}"; do
        if [[ ${Args[$Arg]} == '-msse' || ${Args[$Arg]} == '-D_USE_SSE' ]]; then
                unset "Args[$Arg]"
        fi
done

exec /usr/bin/gcc "${Args[@]}"
```

Also adding -Wno-error=narrowing to the cxxflags to avoid the other issue this PR fixes.  So my Apt repo now publishes an arm64 build.

Not to be impatient, but since I'm adding this info and making another comment anyway: @dleslie do you intend to update your PR or should I create a new one/take over?

Blzut3:
Superseded by merged PR #1544.

PR #1237, opened 2021-02-28
Fixes to flat texture selection

Talon1024:
This improves separation of flats from wall textures, and make flat texture selection work like in GZDoom (AFAIK anyway). This fixes a bug where flat textures weren't scaled if they used a high-res texture.


sirjuddington:
Looks good, the only real issue is it is difficult to merge this into master (3.2.0) due to changes to MapTextureManager in particular. Still I'll merge this in and look at master later heh

PR #1244, opened 2021-04-04
Add the ability to convert to PSX DOOM format textures

BodbDearg:
This pull request adds the ability for SLADE to convert images to the PSX Doom graphic format. SLADE can already read this format very well; all that is required is to extend the capabilities to writing. Most of the code is also copied from similar code for other formats.

This change was motivated by the advent of new tools which enable mapping for PlayStation Doom, and the reverse engineered PC backport, PsyDoom.

Note that this change is primarily intended to help out mappers targeting 'PsyDoom'. For the native PSX engine you would also need to update the `TEXTURE1` lump if adding in new wall textures; PsyDoom plans to bypass use of that lump however to make modding easier. `TEXTURE1` just contains an array of `PSXPicHeader` (one for each wall texture in the game between `T_START` and `T_END`) and allows the engine to know the size of each wall texture before loading it.

### DIFF
--- a/src/Graphics/SImage/Formats/SIFDoom.h
+++ b/src/Graphics/SImage/Formats/SIFDoom.h
@@ -642,6 +642,40 @@ public:
 		return info;
 	}
 
+	Writable canWrite(SImage& image) override
+	{
+		// Must be converted to paletted to be written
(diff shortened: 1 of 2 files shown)


BodbDearg:
Thank you! :)

PR #1251, opened 2021-04-30
Fix fully black pixels being made transparent when converting to the PSX image format

BodbDearg:
The color 0,0,0 is considered fully transparent by the PSX hardware. If opaque black is used in a source image then it must be converted to a color close to black for the PSX format. Failure to do this can result in 'holes' in the image where black is used.

### DIFF
--- a/src/Graphics/SImage/Formats/SIFDoom.h
+++ b/src/Graphics/SImage/Formats/SIFDoom.h
@@ -733,12 +733,16 @@ protected:
 
 	// Helper: corrects for the opaque RGB colour '0,0,0' being considered fully transparent by the PSX hardware.
 	// Swaps it for the next nearest colour to black in the palette, otherwise there may be unwanted holes in the image where black is used.
+	// Black with the semi-transparency flag set will be used as the swap color, if it is found in the palette.
+	// This special PSX color can be used to achieve true black since most geometry in PSX Doom is rendered with opaque draw commands.
+	// For more on this problem, and the 'black with semi-transparency flag' workaround see the 'No$PSX Specifications' under the 
+	// "GPU Video Memory (VRAM) -> Texture Color Black Limitations" section: http://problemkaputt.de/psx-spx.htm#gpuvideomemoryvram
 	static void correctOpaqueBlackForPsx(SImage& image)
 	{
 		if (!image.palette())
 			return;
 
-		const short nearBlackIndex = findClosestNearBlackColourIndex(*image.palette());
+		const short blackColorIndex = getPsxOpaqueBlackColorIndex(*image.palette());
 
 		for (int y = 0; y < image.height(); ++y)
 		{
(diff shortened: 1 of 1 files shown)


Gaerzi:
This has the drawback of making that conversion a lossy one. Maybe put a message box warning about the issue when opaque black pixels are present? (With a "yes to all" option when batch-converting.)

BodbDearg:
Thank you for reviewing. Upon further investigation I think there might actually be a bit of room for improvement with this, which could avoid minor color loss on black.

The 1st color index for the main PSX palette is of course fully transparent (all 16-bits of the palette entry are 0) but I noticed that color index 247 is also 0,0,0 but presumably has the high bit set (semi-transparency flag, 0x8000) in the 16-bit palette entry. Under certain situations and with certain types of rendering commands, it's possible to do fully black if the color is black and but has the 'semi transparency' flag set. For more about this, see the 'No$PSX Specifications' under the "GPU Video Memory (VRAM) -> Texture Color Black Limitations" section:
http://problemkaputt.de/psx-spx.htm#gpuvideomemoryvram

I *think* this trick might be actually possible to use for Doom based on the type of rendering commands used. If this turns out to be true and color index 247 is black with semi-transparency, then I could amend the commit to try and use this color instead. I would have to hardcode the assumption that any instance of black outside of color index '0' is black with semi-transparency, given that SLADE has no concept of the PSX 'semi-transparency' flag in its color data. We're already doing similar hardcoded assumptions on importing the PSX image anyway - in that situation color index '0' is assumed to be transparent, which it will be unless someone drastically rearranges the palette for some reason. These assumptions are a slight cheat around the lack of the semi-transparency flag in SLADE's color model, but it's probably not worth polluting the model with this PSX concept to support a relatively niche format like PSX.

Anyways I will investigate to see if this can be improved further using the trick described above. The main thing I'm trying to avoid is seeing this (holes in walls) after importing a new texture for the PSX engine:
image

There was a few instances of this issue in the original game too (some pixels of the sky can be seen through the metal grate):
image

BodbDearg:
Ok that seemed to work very well, what I was proposing above... Color index 247 is indeed black with semi transparency; you can verify this by examining the little endian 16-bit color value found at offset `494` of PLAYPAL in PSXDOOM.WAD, it is 0x8000. When I used this special 'semi transparent black' color in PsyDoom (which emulates PSX shading) it worked well and no holes in textures were observed.

This should fix the concern over color loss here, since we can now use this special black to losslessly convert.

BodbDearg:
Made another small tweak in 0251df22afb5936ea883e4a88bcc130995f7e62b. I'm now restricting the opaque black fix for when color index '0' is used but when opaque black was intended. All other colors will be left untouched regardless of their values.

BodbDearg:
Thank you @Gaerzi!

Issue #1134, opened 2020-04-29
Lua's "print" function does not work


flarn2006:
**SLADE Version**: 3.1.11
**Operating System**: Linux

**Issue Details**:
When running a Lua script that uses the `print` statement, nothing appears in the output window.

flarn2006:
I just read about `App.logMessage`, so I guess that's what I'm supposed to use, but it's not intuitive for `print` to do nothing, and it also might mislead someone into thinking their code isn't reaching a certain point. So might as well have `print` work as well.

PR #1254, opened 2021-05-09
Change "print" from "logMessage" to "LogMessage"

flarn2006:
All of the functions have apparently been changed to start with capital letters, but `new_print` hasn't been updated to use the new capitalization. Since Lua is case-sensitive, that means issue #1134 has returned. This pull request fixes the issue.

When was this capitalization change made, however? If it's only in the beta for 3.2.0, and not any prior versions, then rather than merging this pull request, I think it would be better to just revert the capitalization change. Unless it serves some important purpose I'm unaware of, it seems like a needless change that breaks compatibility with old scripts for no apparent benefit. It is of course up to you; that's just my advice.

### DIFF
--- a/src/Scripting/Lua.cpp
+++ b/src/Scripting/Lua.cpp
@@ -179,7 +179,7 @@ bool lua::init()
             for i,v in ipairs({...}) do\
                line = line .. tostring(v) .. ' '\
             end\
-            App.logMessage(line)\
+            App.LogMessage(line)\
          end";
 	lua.script(new_print);
 


sirjuddington:
The lua scripting API in 3.1.x was always marked as a WIP with the possibility of large breaking changes in the future, 3.2.0+ will be a lot more stable API-wise.

I changed functions to PascalCase to make them distinct from properties (in camelCase), which should help make it more obvious when to use `:` instead of `.` (I know this doesn't work for functions in namespaces as opposed to types, but there are a lot less of them overall)

Issue #1270, opened 2021-07-11
MacOS support


Pedro-Beirao:
**SLADE Version**: 
**Operating System**: OSX

@sirjuddington 
Hey Im the guy who asked about the UMAPINFO support 3 days ago.
I'd be happy to maintain the macos support, compile it whenever there's a new release.
I'm also the maintainer of the macos support of dsda-doom source port btw

Pedro-Beirao:
So I just tried compiling the latest version of Slade, and I encountered a lot of problems, but I was able to fix them all.
Im going to make a pull request with the necessary changes to make it work on MacOS

sirjuddington:
Yeah thanks for the pr, an active macos maintainer is sorely needed (the last working macos release is nearly 2 years old now)

PR #1271, opened 2021-07-12
Making it work on MacOS

Pedro-Beirao:
As stated on the #1270 issue, I'm fixing MacOS support.

Changes:
1. The **lstdc++** library was deprecated several years ago (Source: https://developer.apple.com/forums/thread/113746)
2. **nil** has to be replaced by **lua_nil** (Source: https://github.com/mamedev/mame/commit/f191d038ec1fa110e06374ff79e345c91b3365d7)
3. **sol.hpp** needs to include *map* and *unordered_map* (Source: https://stackoverflow.com/questions/15733163/c-error-unordered-map-does-not-name-a-type) (It might also work if you only include *unordered_map*)

I also had a few problem with some *#include* statements with the **ftgl** library but thats something that someone trying to compile has to deal with imo.

You should try to compile this to Windows to check if there are no problems, before merging, obviously.

As stated on the issue I opened, Im available to maintain support for MacOS


src/CMakeLists.txt
```diff
@@ -206,7 +206,7 @@ target_link_libraries(slade
 	${LUA_LIBRARIES}
 	${MPG123_LIBRARIES}
 	${fmt_LIBRARIES}
-	-lstdc++fs
+
```
alexey-lysiuk, line 209: Doesn’t it break linking on Linux?

src/Scripting/Export/Archive.cpp
```diff
@@ -39,6 +39,7 @@
 #include "Utility/StringUtils.h"
 #include "thirdparty/sol/sol.hpp"
 
+
```
alexey-lysiuk, line 42: Please remove this useless change.

alexey-lysiuk (changes requested):
Please delete `.DS_Store` files.

src/CMakeLists.txt
Pedro-Beirao: I have not tested on Linux, I was hopping someone would give me feedback about that.
Deleting it is necessary for it to work on MacOS.

src/Scripting/Export/Archive.cpp
Pedro-Beirao: I was testing something and forgot that. I'll fix that and the .DS_STORE files

src/CMakeLists.txt
Pedro-Beirao: Ok, now it check if its macOS or not

sirjuddington:
Compiles fine in windows still, if @alexey-lysiuk doesn't have any other issues with it i'll merge it in

src/CMakeLists.txt
alexey-lysiuk: I cannot check right now, but is it a correct CMake syntax?

Pedro-Beirao:
@alexey-lysiuk you were right about the cmake file. I can't use an *if* statement inside of the *target_link_libraries*
I changed it and it should be fine

src/CMakeLists.txt
```diff
@@ -192,22 +192,40 @@ add_executable(slade WIN32 MACOSX_BUNDLE
...
-	${OPENGL_LIBRARIES}
-	${GLEW_LIBRARY}
-	${CURL_LIBRARIES}
-	${LUA_LIBRARIES}
-	${MPG123_LIBRARIES}
-	${fmt_LIBRARIES}
-	-lstdc++fs
-)
+if(APPLE) 
+	target_link_libraries(slade
```
alexey-lysiuk, line 196: `target_link_libraries()` can be used several times with the same target. Please restore the initial version without `stdc++fs`, and add it for Linux only with a separate call.

Actually, it should be added on any platform if we link with stdc++ standard library implementation.
However, proper detection of filesystem support in general requires a [separate module](https://github.com/vector-of-bool/CMakeCM/blob/6cb8a6ada09eb40c271d2f83a032e03cd4005fa3/modules/FindFilesystem.cmake).

Pedro-Beirao: Done

src/CMakeLists.txt
```diff
@@ -192,22 +192,25 @@ add_executable(slade WIN32 MACOSX_BUNDLE
...
+		${wxWidgets_LIBRARIES}
+		${FREEIMAGE_LIBRARIES}
+		${SFML_LIBRARY}
+		${FTGL_LIBRARIES}
+		${OPENGL_LIBRARIES}
+		${GLEW_LIBRARY}
+		${CURL_LIBRARIES}
+		${LUA_LIBRARIES}
+		${MPG123_LIBRARIES}
+		${fmt_LIBRARIES}
```
alexey-lysiuk, line 209: Please avoid useless changes in spacing like this altered indent and empty line. It’s much harder to review such changes. Git diff and blame features will produce unneeded noise.

sirjuddington:
Closing this one since #1279 replaced it, but thank you @Pedro-Beirao for helping out with the macos stuff

PR #1296, opened 2021-11-11
ArchivePanel: deleted windows should be replaced in the splitter

kevans91:
The sizer contains the splitter, which contains the two panes.  When an
archive entry is deleted, the splitter's the one that we need to be
replacing it in.

This fixes an assertion when an archive entry is deleted, then another one
is selected. This regression seems to be from eb6eeb5db6fa, when the
splitter was introduced.


kevans91:
I guess I omitted: the assertion is specifically about a non-existent window from `splitter_->ReplaceWindow()` in `ArchivePanel::showEntryPanel`.

PR #1297, opened 2021-11-20
Small collection of fixes

kevans91:
- Report glewInit() errors
- Flush log on write so that we don't lose log entries that can clue us into the problem if the application crashes
- Fix recent issue where special args can't be saved


kevans91:
I threw in one more tiny fix to avoid adding invisible prats of lines to the selection when moving from 2D to 3D mode, which fixes deselecting affected lines in 3D mode (e.g., doors, windows, steps)

Issue #1100, opened 2020-01-04
Fix HiDPI support


MarkJeronimus:
**SLADE Version**: (v3.1.7)
**Operating System**: (Ubuntu 18.04 with Unity)

**Issue Details**:
When using a High DPI (HiDPI) setting in the OS (200% in my case) the 3D scene of the map editor is rendered in the lower-left corner of the viewport. This isn't a bug in Slade per-se because I see the _exact_ same problem in another program, KiCAD. The problem in KiCAD can be worked around by disabling anti-aliasing.

However, there's no way (I found) to disable anti-aliasing in the map editor.

https://youtu.be/NGrIPmQM9pw

matatk:
Hi, I believe I'm experiencing the same issue on macOS, using a MacBook Pro from several years back. When running on the internal display, I get exactly the same problem as above. When running on an external monitor (in non-HIDPI mode), SLADE's editor looks as expected. SLADE looks pretty awesome; looking forward to using it :-).

PR #1301, opened 2021-12-15
Add initial support for HiDPI scaling, tested on an M1 Max Macbook Pro

### LINKED ISSUE #1211: Inconsistent text scaling on HiDPI monitor
**SLADE Version**: v3.2.0 beta 2
**Operating System**: Linux
**Issue Details**:
I'm compiling SLADE against GTK3 and WxWidgets v3.0.4 on NixOS. in a HighDPI environment, part of the GUI scales correctly (namely, the dropdown menus), but everything else is too small. You can see an example of how slade looks for me [here](https://imgur.com/a/rgYr4uX). I can get sensible results only by setting the following environment variables; `GDK_SCALE` scales up the text that is too small, although then I have to scale the menu text back down using `GDK_DPI_SCALE`. If I do this, everything generally looks good, but I get bitten by the issue described in #1100 

```
GDK_SCALE=2
GDK_DPI_SCALE=0.5
```
Other GTK3 apps generally scale fine.

MP2E:
fixes #1211 and #1100

Note that I'm not sure if I was able to test everything SLADE has to offer, I simply loaded up DOOM 2 and looked at several different types of lumps and made sure they look ok on my hidpi monitor. I also may have changed things that didn't need changing, I was using grep to find where I needed to make changes based on surrounding functions.

### DIFF
--- a/src/MainEditor/UI/MainWindow.cpp
+++ b/src/MainEditor/UI/MainWindow.cpp
@@ -467,8 +467,9 @@ bool MainWindow::exitProgram()
 	// main_window_layout = aui_mgr_->SavePerspective();
 	saveLayout();
 	mw_maximized = IsMaximized();
+	const wxSize size = GetSize() * GetContentScaleFactor();
 	if (!IsMaximized())
-		misc::setWindowInfo(id_, GetSize().x, GetSize().y, GetPosition().x, GetPosition().y);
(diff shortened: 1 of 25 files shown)


MP2E:
Also, some of the UI elements on the map editor are very small now on HiDPI monitors, I will see if I can fix that too. Here's a picture showing the issue
Screen Shot 2021-12-15 at 12 11 01 AM

PR #1313, opened 2022-02-01
ZDoom Texture Cleanup Operations

ill:
This adds 2 right click options for texture entries and an archive maintenance option for finding all unused zdoom textures.

Screen Shot 2022-02-01 at 1 30 25 AM

Screen Shot 2022-02-12 at 5 49 56 AM

Doom style and Zdoom style textures both have an option to clean out duplicate unmodified texture entries that also exist in the iwad.

Zdoom style textures also have another option to clean out redundant single patch textures and convert those patches into textures themselves.

The Remove Unused Zdoom textures option scans through all wall and floor/ceil textures in every map similar to the existing remove unused textures/remove unused flats tools.  Except remove unused textures currently doesn't check floor/ceil, and remove unused flats currently doesn't check walls.  Also I have a rudimentary Animdefs parser to add animated and switch textures to the exclude list.  This will remove texture/flat images themselves from the archive, and it will remove texture entries from texturex and textures files.

This way you can easily do some large scale cleanup when working with texture packs and making them ready for zdoom format games.

### DIFF
--- a/dist/res/actions/arch.cfg
+++ b/dist/res/actions/arch.cfg
@@ -61,6 +61,12 @@ action arch_clean_flats
 	help_text	= "Remove any unused flats";
 }
 
+action arch_clean_zdoom_textures
+{
+	text		= "Remove Unused &Zdoom Textures";
+	help_text	= "Remove any unused textures, flats, and patches in a zdoom archive.";
+}
+
 action arch_check_duplicates
 {
 	text		= "Check Duplicate Entry Names";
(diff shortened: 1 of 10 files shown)


Gaerzi:
Oh, nice! I hope @sirjuddington will pop in to take a look; he hasn't been very active recently.

ill:
Just added a remove unused zdoom textures archive cleanup operation.

PR #1395, opened 2022-06-19
Add support for Doom 32X map format

viciious:
We go from this:

image

to this:

image

The format is detected by the presence of the "PLAYPALS" lump. which is unique to 32X WADs.


viciious:
Thanks!

PR #1413, opened 2022-08-14
Allow the system fmt and DUMB libraries to be used rather than the bundled ones

chewi:


konchunas:
This is a good change since it allowed me to successfully compile the project on M1 Mac.
Bundled `fmt` failed with all kinds of errors.
With this merged `cmake -DUSE_SYSTEM_FMT .` did the job.
Thank you!

PR #1436, opened 2022-11-12
ZDoom language definitions update

Blue-Shadow:


Blue-Shadow:
More stuff has been added since the PR was made, so I thought I close it and open another one at a later point.

PR #1455, opened 2023-01-28
Fixed a bug in the regex that detected decimal values, which fixes a …

CandiceJoy:
…frequently-occurring crash.


sirjuddington:
This one seems to be doing a lot more than the title says - any chance this could be split into multiple single-issue PRs? Or at least have the DUMB->XMP replacement part split out

Issue #1016, opened 2019-01-05
SLADE performance gradually degrades until it's unusable


eevee:
**SLADE Version**: master, for months
**Operating System**: Arch Linux x64

**Issue Details**:

If I leave SLADE open for a couple hours, even if it's completely idle, it will gradually consume more CPU time and take longer to redraw until it's nigh unusable.  Even closing it is harder, because the "do you want to exit" dialog doesn't draw its contents (despite the window itself opening immediately).

I've thrown everything I can think of at this and have absolutely no idea what's going on.  Both `perf` and manual inspection with `gdb` blame most of the time on an unnamed function a couple calls below `g_main_context_check`, e.g. `perf report` shows:

```
           80.21% 0x7f49cc1926d7
           6.95% 0x7f49cc192677
           4.20% g_main_context_check
           4.12% g_main_context_prepare
           0.88% 0x7f49cc1926f3
```

But this is just part of glib's main loop, which should be `poll()`ing and not eating up any real CPU time.

Doesn't like to reproduce unless I'm not trying to reproduce it, which makes bisecting difficult.  Seems to happen faster if I'm actively using the map editor, but that might be my imagination.  Regardless, SLADE is practically unusable for any real work like this.

Would love some suggestions, because I'm out of ideas.

zgramps:
Not sure if this is helpful, but this only seems to happen when map editor is left open. Have left the main part of Slade open overnight with no problems, but map editor will pretty consistently grind to a near-halt.

eevee:
I want to say it's happened once or twice with only an archive open (and no map editor), but I haven't been keeping super detailed notes or anything.

Gaerzi:
Haven't noticed this problem on Windows, so it's probably Linux-specific.

sirjuddington:
If it is actually linux-only (or non-windows-only) it may be something to do with either the wxGLCanvas implementation of OGLCanvas, or the FTGL font rendering implementation

eevee:
Would early returning from `Drawing::drawText` be enough?

eevee:
Early returning from `Drawing::drawText` did not fix it.  Also it looks like I can reproduce pretty reliably by just leaving the map editor open for 24 hours.  Worst possible debug cycle.  :)

OrdinaryMagician:
Ahhhh I knew I wasn't alone on this. Good catch.

eevee:
[...]

PR #1462, opened 2023-03-21
Fix SLADE on Linux

eevee:
You won't believe how!  It's by turning off word wrap in the console (on wxGTK only; no reason to do it for anyone else).  See #1016 and wxWidgets/wxWidgets#23364.

This is a kludge until...  someone else fixes it for real.  But the drawback is pretty minor, considering I haven't been able to use SLADE for mapping at all for five and a half years.

The rest of this fixes some assertion failures and stderr spam.  Not sure why the map editor window's toolbar buttons are getting mouse events when it's not even open...


sirjuddington:
Oh wow, that's... obscure :P

Thanks for all the work you put into investigating this one, hopefully it'll be fixed in wxGTK soon

eevee:
They do have a patch already, but it's simmering on master and no telling when it'll be in a 3.2 release.  (And SLADE doesn't currently compile against 3.3.)

PR #1501, opened 2023-05-21
dereference null pointer directly in crash command

Jan200101:
`test[123]` would result in `*(test+123)`, which may point to valid memory that we modify. Instead dereference the null pointer directly to ensure we actually crash

### DIFF
--- a/src/Application/SLADEWxApp.cpp
+++ b/src/Application/SLADEWxApp.cpp
@@ -771,7 +771,7 @@ CONSOLE_COMMAND(crash, 0, false)
 		== wxYES)
 	{
 		uint8_t* test = nullptr;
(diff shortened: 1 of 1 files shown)


src/Application/SLADEWxApp.cpp
```diff
@@ -771,7 +771,7 @@ CONSOLE_COMMAND(crash, 0, false)
 		== wxYES)
 	{
 		uint8_t* test = nullptr;
-		test[123]     = 5;
```
Jan200101, line 774: Going a bit further, this could be improved by making it volatile, since the compiler may decide that this is nonsense and to ignore it.

`abort()` may even be better, but that results in a different reror.

PR #1561, opened 2023-08-30
Replace deprecated wxBitmap methods (fix build on wxWidgets 3.3)

joanbm:
Stop using some deprecated `wxBitmap` methods (`wxBitmap::SetWidth`, `wxBitmap::SetHeight`) which have been deprecated since
wxWidgets 3.1.2 and are removed in the current v3.3 development trunk.

With this SLADE builds fine on the current wxWidgets 3.3 development trunk.
I also verified the changes against wxWidgets (wxGTK) 3.0.5 and 3.2.4.

### DIFF
--- a/src/TextEditor/UI/SCallTip.cpp
+++ b/src/TextEditor/UI/SCallTip.cpp
@@ -60,7 +60,8 @@ CVAR(Bool, txed_calltips_dim_optional, true, CVar::Flag::Save)
 // -----------------------------------------------------------------------------
 // SCallTip class constructor
 // -----------------------------------------------------------------------------
-SCallTip::SCallTip(wxWindow* parent) : wxPopupWindow(parent), buffer_{ 1000, 1000, 32 }, font_{ GetFont() }
+SCallTip::SCallTip(wxWindow* parent) : wxPopupWindow(parent),
+	scratch_{ 1000, 1000, 32 }, buffer_{ 1, 1, 32 }, font_{ GetFont() }
 {
 	wxPopupWindow::Show(false);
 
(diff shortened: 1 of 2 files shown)


sirjuddington:
Unfortunately this breaks tooltips in windows. It looks like there is some kind of bug with the wxBitmap copy constructor (unless there's something I'm missing) - the line `buffer_ = scratch_.GetSubBitmap(...` makes `buffer_` just an empty black bitmap for me on wxMSW 3.2.2.

I managed to get around the issue by removing `buffer_` altogether, keeping a size member variable and using `scratch_.GetSubBitmap` directly with the size when drawing the tooltip. Another alternative is to do the same thing but just draw `scratch_` at full size directly without `GetSubBitmap`, which avoids creating a new bitmap each time it's redrawn. I suppose with this workaround `scratch_` could also just be renamed back to `buffer_` :P

joanbm:
There's indeed something going wrong on Windows - I'll take a look...

joanbm:
FWIW, the fact that `wxBitmap::GetSubBitmap` returns a black bitmap isn't due to a wxWidgets bug but rather due an API misuse in the PR - [the wxBitmap must be selected out of the wxMemoryDC before reading the image data](https://docs.wxwidgets.org/3.2/classwx_memory_d_c.html#details).

joanbm:
PR updated, Windows works now.

I kept the same approach of using `wxBitmap::GetSubBitmap`, but propely managing the lifetime of the DC associated to the bitmap now.

PS: I also dropped the part of the PR fixing `SToolBarGroup::addActionGroup` since that was already done by 99d044ef0e7ea0887470a7aa0667f026eb727d18.

PR #1591, opened 2023-11-05
Only check the MSB of the Jaguar sprite post for 0xFF

viciious:
This matches behavior of Calico Doom's renderer and also adds support for lumps that contain unaligned column data, where a single post byte of 0xFF is used a post:
https://github.com/team-eternity/calico-doom/blob/master/src/r_phase8.c#L39


viciious:
No idea what SonarCloud's problem is

PR #1594, opened 2023-11-06
Extend the Jaguar image format and add conversion option

viciious:
I would like to use the Jaguar image format for wall textures for the next version of Doom 32X Resurrection, hence the change. The "standard" texture format in JagDoom is rather silly since there's no header at all and the only way to auto-detect it is by checking if the 320-byte footer in the lump matches the initial 320 bytes. With 100 textures, that would require roughly 32KB of ROM space, which is just too wasteful, since the data in the footer is redundant to begin with.

So my plan is to re-use the JaguarImage format used for 2D gfx for wall textures, but also extend it by adding two new flags in the unused padding space of the header:
- bitflag of 1 indicates the column-major format

Before:
Screenshot from 2023-11-06 21-04-20

After:
Screenshot from 2023-11-06 21-03-06

### DIFF
--- a/dist/res/config/entry_types/gfx_doom.cfg
+++ b/dist/res/config/entry_types/gfx_doom.cfg
@@ -146,6 +146,17 @@ gfx_doom_jaguar : gfx_base
 	reliability = 65;
 }
 
+gfx_doom_jaguar_colmajor : gfx_base
+{
+	name = "Graphic (Jaguar CM)";
+	format = img_doom_jaguar_colmajor;
+	editor = gfx;
+	extra = image;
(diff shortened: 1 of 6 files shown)


viciious:
As a bonus, the auto-detection of 32x16 JagDoom textures is fixed in commit 22db06f3723a08e7f2f1d7cffc1dd3e3554afe1f

viciious:
I've also added conversion to the aforementioned format, however I don't know how I should go about the column-major subformat.

viciious:
image

I've slightly rewritten the code to allow exporting images to the column major format.

PR #1597, opened 2023-11-11
Fix mirroring actions to only apply once with keyboard shortcut

thezerobit:
I noticed that if I use the keyboard shortcuts in 2D map edit mode to mirror selection, it would apply the action twice. You have to use Undo to see that it was mirrored twice, otherwise it looks like nothing happened. This doesn't happen when you use the menu. The action in either case is triggered from the MapEditContext::handleAction method, so this code in Input::handleKeyBind2d is superfluous and just causes it to be applied twice when using the shortcuts. After removing these lines, the keyboard shortcuts became usable, just triggering the action once.

### DIFF
--- a/src/MapEditor/Edit/Input.cpp
+++ b/src/MapEditor/Edit/Input.cpp
@@ -896,12 +896,6 @@ void Input::handleKeyBind2d(string_view name)
 				context_.addEditorMessage("Selection numbers disabled");
 		}
 
-		// Mirror
-		else if (name == "me2d_mirror_x")
-			context_.edit2D().mirror(true);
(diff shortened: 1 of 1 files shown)


sirjuddington:
Closing this PR as I've fixed a different way in e542d0dabdec3f37f588453d85404a87f627977c, which should also work for any other actions that have a similar setup (or have their keybinds changed by the user to something with `ctrl` or `alt`)

Interestingly this didn't happen on Windows, as there for some reason if a keypress hits a menu item with a shortcut key, it isn't propagated further and so doesn't get to the keybind system.

Issue #1112, opened 2020-02-25
Unable to use Mac Version


VGBoy:
**SLADE Version**: (3.1.10 (Reads as 3.1.9-4-ga5927973 on the dmg itself)
**Operating System**: (OSX Version 10.14.6)

**Issue Details**:
When I try to open up SLADE, It displays this message and forces me to delete it if I copy it to Applications or eject the disk image while it's still in the dmg.
<img width="415" alt="Screen Shot 2020-02-24 at 10 28 24 PM" src="https://user-images.githubusercontent.com/18492821/75212503-21da1b00-5755-11ea-875c-0d6e61ceabc1.png">
I don't know if it is just the download being corrupted or there is actually something wrong with the Mac version specifically.

gernotresch:
having the issue as well. copying the binary out of the image doesn´t solve the problem.
macOS 10.15.3 here.

sirjuddington:
No idea what is wrong with it, does 3.1.9 work?

VGBoy:
The old 3.1.9 download works with no issue.

gernotresch:
for me as well.

sirjuddington:
I made a build for [3.1.11](https://github.com/sirjuddington/SLADE/releases/tag/3.1.11), does that work?

VGBoy:
I've tried opening the 3.1.11 build. It still says the app is damaged and can't be opened.

sirjuddington:
Hmm well that's annoying. Unfortunately I have no idea what could be wrong with it and don't really have the mac expertise to figure it out

gernotresch:
ah that´s sad. but nevertheless, thanks for your efforts!

alexey-lysiuk:
Running `pkgutil --check-signature /Volumes/SLADE-3.1.11/SLADE.app` in Terminal outputs `Status: package is invalid (checksum did not verify)`
Most likely, it wasn't singed with the certificate trusted by Apple.

matatk:
I've been unable to use 3.1.11 for the same reasons and, oddly, I thought I'd got my copy of 3.1.9 from the GitHub Releases page, but it doesn't work for the same reason.

I installed wxWidgets and tried to compile (both `master` and `stable`) but get this error, which I've not been able to resolve so far...

```
Could NOT find wxWidgets (missing: wxWidgets_LIBRARIES
wxWidgets_INCLUDE_DIRS) (Required is at least version "3.0.4")
```

[...]

PR #1628, opened 2023-12-18
Request Accessibility Permissions on Mac

Pedro-Beirao:
The 3d mode camera spins like crazy if this permission is not given to Slade. So this shows a Pop Up that tells the user how to fix it

Before this, users needed to google or ask someone how to fix this problem

### DIFF
--- a/src/Application/App.cpp
+++ b/src/Application/App.cpp
@@ -62,6 +62,9 @@
 #include "Utility/Tokenizer.h"
 #include <dumb.h>
 #include <filesystem>
+#ifdef __WXOSX__
+	#include <ApplicationServices/ApplicationServices.h>
+#endif
 
 using namespace slade;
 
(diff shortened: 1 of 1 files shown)


Pedro-Beirao:
Also, I would like to make a Mac build to update the one in the website, since that one is pretty old (2019)

What do you think?

sirjuddington:
If you can make a redistributable build (eg. dmg like the existing ones) for 3.2.4 that would be awesome, I haven't been able to for years and probably still won't for the forseeable future.

Also note I'll probably be releasing 3.2.5 sometime soon so not sure if it would be best to wait until that is finalised (though feel free to see if you can get one going for 3.2.4 in the meantime)

I used to use https://github.com/alexey-lysiuk/slade-macos-build to make the builds though it looks like it's been archived now and is very out of date so I doubt it would still work. There is also issue #1112 which is one of the main reasons the mac build is still stuck at 3.1.6.

Pedro-Beirao:
I dont need that repo for making the build.
The issue linked seems to be due to unsigned binaries. I will be codesigning it, but not with an Apple Developer account since I dont currently have it
(People will need to right click -> open)

I'll probably contact you through discord when I get this done

PR #1644, opened 2023-12-31
CMake cleanup and fixes

### LINKED ISSUE #1641: The `sudo make install` / `sudo ninja install` fails because of `slade.pk3` file missing
**SLADE Version**: top of branch `master`
**Operating System**: Linux (Ubuntu 22.04)

**Issue Details**:
Recently, I finally figured how to properly install dependencies and build the thing. However, it fails to be installed because of the `slade.pk3` file is missing. Is that supposed it being no longer generated?

The log I get on attempt to install the thing:
```
$ sudo ninja install
[0/2] Re-checking globbed directories...
[0/1] Install the project...
-- Install configuration: "Release"
-- Installing: /usr/local/bin/slade
CMake Error at src/cmake_install.cmake:67 (file):
  file INSTALL cannot find
  "/home/vitaly/_git_repos_other/SLADE/dist/slade.pk3": No such file or
  directory.
Call Stack (most recent call first):
  cmake_install.cmake:47 (include)


FAILED: CMakeFiles/install.util 
cd /home/vitaly/_git_repos_other/SLADE/dist && /usr/local/bin/cmake -P cmake_install.cmake
ninja: build stopped: subcommand failed.
```

SteelT1:
- Replace calls to deprecated exec_program with execute_process when using uninstall target
 - Add comment for uninstall target
 - Only install slade.pk3 if BUILD_PK3 is enabled (fixes #1641)

### DIFF
--- a/cmake/cmake_uninstall.cmake.in
+++ b/cmake/cmake_uninstall.cmake.in
@@ -7,10 +7,10 @@ string(REGEX REPLACE "\n" ";" files "${files}")
 foreach(file ${files})
   message(STATUS "Uninstalling $ENV{DESTDIR}${file}")
   if(IS_SYMLINK "$ENV{DESTDIR}${file}" OR EXISTS "$ENV{DESTDIR}${file}")
-    exec_program(
-      "@CMAKE_COMMAND@" ARGS "-E remove \"$ENV{DESTDIR}${file}\""
+    execute_process(
+      COMMAND "${CMAKE_COMMAND}" -E remove "$ENV{DESTDIR}${file}"
       OUTPUT_VARIABLE rm_out
-      RETURN_VALUE rm_retval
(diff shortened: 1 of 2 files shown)


Pedro-Beirao:
To add to this, any reason BUILD_PK3 is not ON by default?

sirjuddington:
I turned it off by default as I was getting sick of it rebuilding the entire pk3 every time I built SLADE during development (since I'm using cmake in windows now instead of the VS project/solution). It could probably be turned on by default if there was a way to have it only rebuild the pk3 if any files changed in dist/res but I'm not sure if that's possible

Wohlstand:
You may want to set the tree of files as dependencies, and have the pk3 build to be a target with pk3 file as a byproduct. I have in my projects a similar thing but with different purposes, but sense is same: source a pile of files and final file as a result of process.

PR #1664, opened 2024-02-09
Enable BUILD_PK3 by default

charlesthobe:
I enabled archive sync so 7z and zip would only touch the pk3 if it differs to the files in `res/`
[This documentation](https://7-zip.opensource.jp/chm/cmdline/switches/update.htm) was used and tested.

Explanation for the option `-up0q0r2x2y2z1w2`:
When updating the archive using 7z, 7z keeps old files by default, so I needed to change that.
Also the documentation is only explaining what happens when creating an update.zip which has the differences not when working on the archive that we want to update directly.
Thus when only the archive that needs to be updated is passed the meaning of the numerical options changes becoming:
0 = Delete
1 = Keep
2 = Update
~~3 = Create antimatter~~ ( not used here :stuck_out_tongue: )

p0 = Delete files from the archive if multiple files were targeted — e.g explicitly or by using a wildcard `*` — without said files being among them.

q0 = Delete files from the archive if not found in target directory (both p and q are similar, one for the former is for multiple targets or a file, the latter is for if the target is a single directory)

r2 = Add files found in target/s and not found in the archive.

x2 = If a file in the archive is newer than in the target, replace it regardless.

y2 = If a file in the archive is older than in the target, replace it.

z1 = If a file in the archive matches that in the target, then keep it (don't do anything)

w2 = If for any reason 7z couldn't figure out if a file matches, then replace it.


charlesthobe:
and for Info-ZIP aka `zip` it's just `-FS` or `--filesync`
it does all of the above.

charlesthobe:
Wait a sec, while zip doesn't touch the archive if there were no changes, 7zip still recreates the archive as slade.pk3.tmp, then deletes the original file and replaces it with the new file :man_facepalming:

charlesthobe:
I opted to create a cmake script instead of integrating the changes into the `CMakeLists.txt` so the command could be run in the build phase instead of the cmake configuration/cache generation, thus it won't be needed to regenerate cmake cache each time a file is changed.

PR #1727, opened 2024-09-23
SFML migration update

jopadan:
[SFML/migration.md](https://github.com/SFML/SFML/blob/master/migration.md) updated with `#if (SFML_VERSION_MAJOR > 2)`


sirjuddington:
This one will probably need some SFML version checks to keep support for 2.x, at least for a while

Blzut3:
The changes to .gitingore here indicate that you're using CMake wrong.  You should set your build directory to an empty directory.  Commonly people just `mkdir build` inside their clone and use that, but it can be entirely out of tree too.  This is important since CMake doesn't provide a way to cleanup its generated files besides "delete the build directory."

sirjuddington:
@jopadan sorry I haven't got to this one yet, could you remove the changes to .gitignore (see @Blzut3's comment) and I'll merge it in

PR #1753, opened 2024-12-08
Add support for the k8vavoom vWad archive format

dashodanger:
The [vWAD](https://repo.or.cz/k8vavoom.git/blob_plain/HEAD:/libs/core/libvwad/vwadvfs.h) format has been in use by k8vavoom for some time now, primarily for its base packs and savegames but it can also be used for external mapsets and mods. EDGE-Classic has also added support for the format to supplement regular zip/pk3/epk/loose folder support. At the moment, there is no real tooling for the format outside of a handful of CLI programs that must be compiled on one's personal box. This adds support for the following vWAD features in the hopes that users of k8vavoom, EDGE-Classic, and possibly ports like Zandronum that have a similar level of DECORATE support as k8vavoom can easily view/edit/create files in this format:

- New vWad creation
- Loading/Editing/Saving existing vWads
- Viewing vWad-specific metadata if present (Author, Title, Comment, Public Key if signed)
- Generating new Private/Public vWad signing key pairs
- Deriving the vWad Public key from a provided Private Key
- Automatically signing saved vWads with a specific key (the 'vwad_private_key' CVAR; if this is blank vWads saved by SLADE will not be signed)
- Automatically setting the 'Author' metadata field for vWads (the 'vwad_author_name' CVAR)

### DIFF
--- a/dist/res/config/archive_formats.cfg
+++ b/dist/res/config/archive_formats.cfg
@@ -43,6 +43,17 @@ archive_formats
 		}
 	}
 
+	vwad
+	{
+		name = "VWad";
+		supports_dirs = true;
+		names_extensions = true;
+		entry_format = "archive_vwad";
+		create = true;
+		allow_duplicate_names = false;
+		extensions { vwad = "k8Vavoom vWAD"; }
+	}
+
 	adat
 	{
 		name = "Anachronox DAT";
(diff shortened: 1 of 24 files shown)


dashodanger:
Did not realize upstream vwadfs libs did not support MacOS until this CI tried to run; will go ahead and close as I will not be trying to maintain a custom fork of the library

Issue #1709, opened 2024-08-17
New text lumps from official re-re-release


Gaerzi:
### Editor

Resource editor

### Description of your feature request

Here's a quick list of lumps that can be found:
- DECOHACK: a DECORATE-like lump containing source code for complex DEHACKED project. This is actually from [MTrop's DoomTools](https://mtrop.github.io/DoomTools/).
- [COMPLVL](https://doomwiki.org/wiki/COMPLVL): a very simple text lump, that actually originates from DSDA-Doom.
- GAMECONF: a JSON-format file with metadata about game configuration.
- SBARDEF: a JSON-format file with a status bar definition.
- SKYDEFS: a JSON-format file with sky definitions.
- XWINTER0 and XWINTER1: JSON-format files with intermission definition. Those cannot be identified by name, though, the lumps are found in found in UMAPINFO "enteranim" and "exitanim" properties, so parsing UMAPINFO to find such lumps is needed.
- TRAKINFO: a JSON-format file which contains SHA-1 hashes of music tracks and associate them to alternative lump names, for the "MIDI" and "Remix" soundtrack options of the new port.
- DMXMIDI: an OPL bank alternative to GENMIDI. Not in the same format as GENMIDI, though. I think it's in [this format](https://github.com/Wohlstand/OPL3BankEditor) but I haven't checked that.
- __VER__: Versioning info in the form of a timestamp
- Text files with extensions in Common.kpf: .json, .kfont (some KEX-engine character sheet definition),  .html, .css, .js, and .js.map and .css.map (HTML/CSS/JavaScript stuff for the game select screen), .shader and .inc (shader source code).
- Binary files with extensions in same: .pat (GUS patches), .ttf (TrueType fonts), .bin and .spirv (compiled shaders, though .bin is a generic extension), .bnvib (controller rumble data).

Text files should definitely be identified properly. Binary files are a lot less important, since after all they can't be edited from within SLADE, and for most of them it'd only be relevant for people looking at the content of Common.kpf.

### Mock-up

![DESCRIPTION](LINK.png)

Blue-Shadow:
COMPLVL is already [recognized](https://github.com/sirjuddington/SLADE/commit/e9207987791898c506e8dc48da22347f6275bcc7).

electricbrass:
Another lump from the new release that isn't on this list is DEMOLOOP. It's another JSON lump that specifies which lumps contain demos to play and graphics to display from the title screen.

PR #1755, opened 2024-12-18
Add support for compiling DECOHack source

electricbrass:
Addresses one part of #1709

DECOHack is seeing a lot of use now for creating DeHackEd patches, and it would be very useful to be able to compile it without leaving SLADE. I've added the ability to point SLADE to the location of a DoomTools jar (and Java executable) as well as options for compiling. Compiling a DECOHACK lump will insert the output as a DEHACKED lump in the wad. I've also added a language definition for syntax highlighting and function tooltips.
image
image
spawnobject


sirjuddington:
Looks good to me, just one minor thing - I think the DECOHack preferences tree item makes more sense to be a child of the 'Scripting' one, since it's pretty much a scripting language + compiler like ACS.

electricbrass:
I was a little unsure on whether it should go there or not. Made the change.

Issue #1759, opened 2024-12-24
Segfault when clicking on music/sound


jiffygist:
### SLADE Version

be9fb027ef1f9df47d365ebdf65b75be456a2fc9

### OS

Linux

### Editor

Resource editor

### Steps to Reproduce

I compiled SLADE on Chimera Linux on Raspberry Pi against latest SFML from git (3.1.0). It crashes when I click on any music or sound.

### Crash Information

```text
Thread 1 "slade" received signal SIGSEGV, Segmentation fault.
0x0000fffff2fc596c in std::__1::pair<std::__1::__hash_iterator<std::__1::__hash_node<sf::Sound*, void*>*>, bool> std::__1::__hash_table<sf::Sound*, std::__1::hash<sf::Sound*>, std::__1::equal_to<sf::Sound*>, std::__1::allocator<sf::Sound*> >::__emplace_unique_key_args<sf::Sound*, sf::Sound* const&>(sf::Sound* const&, sf::Sound* const&) () from /usr/local/lib/libsfml-audio.so.3.1
(gdb) bt
#0  0x0000fffff2fc596c in std::__1::pair<std::__1::__hash_iterator<std::__1::__hash_node<sf::Sound*, void*>*>, bool> std::__1::__hash_table<sf::Sound*, std::__1::hash<sf::Sound*>, std::__1::equal_to<sf::Sound*>, std::__1::allocator<sf::Sound*> >::__emplace_unique_key_args<sf::Sound*, sf::Sound* const&>(sf::Sound* const&, sf::Sound* const&) () at /usr/local/lib/libsfml-audio.so.3.1
#1  0x0000fffff2fc4f48 in sf::SoundBuffer::attachSound(sf::Sound*) const () at /usr/local/lib/libsfml-audio.so.3.1
#2  0x0000fffff2fc286c in sf::Sound::Sound(sf::SoundBuffer const&) () at /usr/local/lib/libsfml-audio.so.3.1
#3  0x0000aaaaab0775e4 in slade::AudioEntryPanel::AudioEntryPanel(wxWindow*) ()
#4  0x0000aaaaab06c314 in slade::ArchivePanel::audioArea() ()
#5  0x0000aaaaab066184 in slade::ArchivePanel::openEntry(slade::ArchiveEntry*, bool) ()
#6  0x0000aaaaab05e77c in slade::ArchivePanel::selectionChanged() ()
[...]

PR #1761, opened 2024-12-24
Adjustments for finding SFML

Blzut3:
As noted in https://github.com/sirjuddington/SLADE/issues/1759#issuecomment-2561479711, it's reasonable to expect that distros are shipping a version of SFML new enough to have the config module in place.  Therefore we can remove FindSFML.cmake.  I see that you were already using the config module on Windows so this ultimately normalizes both sides there.

I also noticed that SFML 3 changed the target names and capitalization of components so it's not backwards compatible.  I've added a bit to try SFML 2 and if not found try with SFML 3 names.  Note that I went this direction due to #1759 indicating that SFML 3 support may need time to bake.  It marginally simpler (one line, one less variable) to prefer SFML 3 if you want to go that route.


sirjuddington:
I think for now we prefer SFML 2, likely will be a while until 3 is more generally available anyway.

Issue #486, opened 2016-03-19
Other minor auto-align quibbles


eevee:
* [ ] There's no way to auto-align to the _left_, which I sometimes want to do to keep the right edge of a texture like `METAL2` aligned with a switch or doorframe or whatever.

* [ ] I'd love to be able to do a "justified" auto-align — given a contiguous set of lines, ensure the leftmost line starts at the left edge of the texture, and the rightmost line ends at the right edge of the texture.  If the total length isn't an exact multiple of the texture's width, spread the leftover space out evenly so there's no obvious seam.

    How good this looks depends heavily on the texture and architecture, but that basic algorithm is something I've done manually a good few times.  Imagine a marble column drawn with the shape tool whose perimeter ends up as being 200 — you can misalign by one pixel at a few vertices, and the obvious 8-pixel overlap all but disappears.

* [ ] Once I've done something like the above, I can't protect all my carefully-arranged offsets from being destroyed by an accidental auto-align.  :)

Gaerzi:
In DB2 and GZDB, autoaligning is done relative to the highlighted sidedef. So if you have the cursor on the leftmost sidedef of a texture span, then all those to the right of it will be aligned to the left. SLADE 3 doesn't do that?

eevee:
No, I think SLADE only aligns to the right.  You're right, though, it would make more sense to just align in both directions automatically.

In which case it might be nice to detect a loop of textures and automatically "justify".

Speaking of autoalign, I swear I've encountered an odd bug a couple times — I'll have a "cavey" room with the same texture all the way around (as confirmed by a shift-click), and I'll press Ctrl-A somewhere on it, but the alignment never makes it all the way around the room to affect the wall to the left.  I've never looked too much into it, though.

PR #1780, opened 2025-02-05
Auto-align in both directions

ralfgerlich:
Current auto alignment only aligns to the right, but not to the left (as lamented in #486). This PR introduces bi-directional auto-alignment for the X-axis.

### DIFF
--- a/src/Application/Main.h
+++ b/src/Application/Main.h
@@ -30,6 +30,7 @@ using std::string_view;
 using std::unique_ptr;
 using std::vector;
 using std::weak_ptr;
+using std::set;
 
 // A macro to check if a value exists in a vector
 #define VECTOR_EXISTS(vec, val) find((vec).begin(), (vec).end(), val) != (vec).end()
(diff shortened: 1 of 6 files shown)


ralfgerlich:
Just realised that the underlying assumption is wrong according to which only walls adjoining the same sector as the initial wall have to be aligned. It's a small fix, but I don't know whether I'll have time before next week.

ralfgerlich:
I have extended the pull-request to jump across two-sided lines. For that I introduced a few helper methods in MapSide.

There is a small open issue in `MapSide::nextInSector` and `MapSide::prevInSector`: Any sector should have at least two sides, so these functions should be able to return something other than `nullptr`. However, can we be certain that the map we're working with actually fulfills this condition? So far, I err on the side of caution by returning a `nullptr` in that case and checking for `nullptr` before application, so it should work regardless even with faulty maps, but I'd rather report a warning here.

Should I use `log::warning`, or should there be checks when loading the map?  What's the right(TM) way to handle this according to SLADE standards?

ralfgerlich:
Closing as it is superseded by #1788.

PR #1819, opened 2025-06-04
Make png optimization use Slade's cache directory.

Person-Who-Exists:
On UNIX, SLADE tried to write PNG files to the location of the optimization tools which is usually located in a directory where regular users don't have write privileges. This pr makes these PNG files get written to SLADE's cache directory.
This pr also fixes 1 miss-spell in the README.
I did not manage to test this pr on windows because I couldn't figure out how to compile SLADE on windows, although it should work because it changes one file path to a path that the rest of SLADE uses.


Person-Who-Exists:
I should add here that this is my first time using c++ so my whole +6 -3 changes of c++ I made might not be that great.

PR #1866, opened 2025-10-21
doc: specify build requirements more exactly

jengelh:
SLADE uses wxBitmapBundle, first introduced with wx 3.1.6. SLADE uses mpg123_ssize_t, first introduced with mpg123 1.28.1.

### DIFF
--- a/COMPILE.md
+++ b/COMPILE.md
@@ -2,13 +2,14 @@
 
 ### Mandatory build-time requirements
 
-* C++17 compiler (e.g. g++ 8.x)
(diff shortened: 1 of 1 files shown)


Blzut3:
It might compile on 3.1.6, but there are known issues solved by newer versions of wxWidgets.  Off the top of my head, at least for Linux, I know anything less than 3.2.3 is known to be buggy.  Really though its been enough of a problem that I'd strongly recommend building with 3.3.1 or if one insists on using a "stable" version 3.2.8.1.

Issue #1595, opened 2023-11-10
Spinning in 3D preview is happening again


JPLeBreton:
**SLADE Version**: 3.2.4
**Operating System**: Linux (Manjaro, stable branch)

**Issue Details**:
Steps to reproduce:
1. Open a map in Slade's map editor
2. Press Q to enter the 3D preview mode
3. Observe: the view is looking straight down at the ground and spinning (rolling). Mouse and keyboard input can change the orientation but the spinning can't be stopped, making it impossible to work in the 3D view.

I know this has been reported and [fixed](https://github.com/sirjuddington/SLADE/issues/1048) a couple times already in the history of the project, but it's started happening again. I'm on a Linux system with an Nvidia GPU (535 drivers) on a GNOME desktop with the Wayland display server. Every other aspect of Slade runs flawlessly, including the 2D map editor view.

eGax:
That issue was referencing MAC builds with spinning issues though.

JPLeBreton:
> Looks like this is another one of those Wayland issues - apparently mouse pointer warping (which 3d mode mouselook relies on to work) is not implemented at all in Wayland (see [wxWidgets/wxWidgets#23778](https://github.com/wxWidgets/wxWidgets/issues/23778)). Not sure if I can do anything about it until it's added unfortunately, as I don't think there is any other way to get mouselook to work using wxWidgets.

Thanks for looking into this, I'll see if there's anything I can do to coax the support along upstream.

iamtoaster:
Maybe it is possible to implement this using GTK directly with some Wayland protocol interactions? I made a proof of concept in #1801 that implements exactly that. It adds an optional dependency on wayland libraries though, and make dependency on GTK3 explicit. I am unsure if it will even work with GTK2. I also did not really think much about how it should fit into existing codebase, so this will definitely need a lot of love. But it works. Kinda. My POC makes the camera movements in 3D view jerky, and I am unsure if it is supposed to be like that. @sirjuddington

Blzut3:
Camera movements are smooth for me under XWayland.  (Probably goes without saying but this is without your change set applied.)

Person-Who-Exists:
There is a new Wayland protocol in the works that would enable cursor warping which should fix this issue.

https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/337

JPLeBreton:
[...]

PR #1892, opened 2025-11-02
Use wxWidgets to center the mouse on the 3D map editor

joanbm:
Upcoming wxWidgets releases will use the `wp_pointer_warp_v1` Wayland protocol for `wxWindow::WarpPointer`, which allows the 3D map editor to work on Wayland compositors that support it (e.g. Plasma 6.5+).

In the mid-long term, this should fix issue #1595.

### DIFF
--- a/src/MapEditor/UI/MapCanvas.cpp
+++ b/src/MapEditor/UI/MapCanvas.cpp
@@ -116,9 +116,9 @@ void MapCanvas::draw()
 // -----------------------------------------------------------------------------
 void MapCanvas::mouseToCenter()
 {
-	auto rect   = GetScreenRect();
 	mouse_warp_ = true;
-	sf::Mouse::setPosition(sf::Vector2i(rect.x + int(rect.width * 0.5), rect.y + int(rect.height * 0.5)));
+	const wxSize size = GetSize();
+	WarpPointer(int(size.x * 0.5), int(size.y * 0.5));
 }
 
 // -----------------------------------------------------------------------------


joanbm:
I did various tests on Windows 11 / Windows 10 / GNOME / KDE / i3 / Sway and wasn't able to find any regression, though I would appreciate more testing. [Specially on MacOS due to this note](https://docs.wxwidgets.org/3.2/classwx_window.html#ac500152d1eca3a2ee98a68e7ea7372b5):

> Apple Human Interface Guidelines forbid moving the mouse cursor programmatically so you should avoid using this function in Mac applications (and probably avoid using it under the other platforms without good reason as well).

sirjuddington:
I doubt we'll get any macOS testing outside any official release, as far as I know that guideline has always been a thing and on macOS SLADE will pop up a dialog to enable permission for it:
https://github.com/sirjuddington/SLADE/blob/79b5da871014356b3dab50682e414c61f5dfa375/src/Application/App.cpp#L584-L597

So I think this should be fine to merge

joanbm:
I realized I didn't originally test the case with scaled displays ("HiDPI"), and found that it regresses e.g. in Plasma X11 with 2x scaling.

(The pointer is actually centered now, but [the camera logic](https://github.com/sirjuddington/SLADE/blob/69b3c7c7da029ef89e08b38bb1950fa7deecc046/src/MapEditor/UI/MapCanvas.cpp#L160-L175) expects it to be off-center).

I'll take a look to see if I can fix forward.

PR #1898, opened 2025-11-04
Add support for base udmf spec

elf-alchemist:
Drafting, as I've come across a minor issue -- in UDMF maps that use the classic style of line specials, the action's target ID is provided on arg0, for all specials (with the exception of the EDGE Classic namespace, that uses only the "Line ID" field). Is there some way to change the default name for arg0, in this case, without needing to update _all_ classic line specials?

Couple test maps in the "Doom" UDMF namespace: [udmf_test.zip](https://github.com/user-attachments/files/23324685/udmf_test.zip)

image

### DIFF
--- a/dist/res/config/games/doom1.cfg
+++ b/dist/res/config/games/doom1.cfg
@@ -10,7 +10,7 @@ game doom
 	iwad = "doom.wad";
 
 	// Supported map formats
-	map_formats = doom;
+	map_formats = doom, udmf;
 
 	// DECORATE filters
 	filters = "doom", "any";
@@ -20,6 +20,9 @@ game doom
 
 	// Light levels
 	light_level_interval = 16;
+
+	// UDMF namespace
+	udmf_namespace = "doom";
 
 	// Maps
 	maps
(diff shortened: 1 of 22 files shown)


sirjuddington:
Looks good so far, however it might be best to do this one on the 3.3.0 branch, as the game configurations there are in a different JSON-based format and organised a little differently. Converting these changes over to the new format would be tricky - small changes should be ok but larger ones like this have an increased risk of something being missed in the conversion, since it would essentially have to be re-done in the new format.

No current way to change the default name of `arg0` but might be possible to add

elf-alchemist:
> it might be best to do this one on the 3.3.0 branch, as the game configurations there are in a different JSON-based format and organised a little differently.

I'll definitely take a look!

> No current way to change the default name of `arg0` but might be possible to add

I believe this will is a crucial UI/UX detail to support, there'll be some definite user confusion, otherwise -- so I will immensely appreciate it if/when it gets added in. :pray:

sirjuddington:
> I believe this will is a crucial UI/UX detail to support, there'll be some definite user confusion, otherwise -- so I will immensely appreciate it if/when it gets added in. 🙏

Yeah, probably will look at implementing it for 3.3.0

elf-alchemist:
Will redo after 3.3.0 is merged

PR #1902, opened 2025-11-09
Fix center check for fractionally scaled 3D map editor

joanbm:
In previous SLADE releases (e.g. 3.2.8), `MapCanvas::mouseToCenter` didn't target the actual center point for scaled ("HiDPI") displays. This is [because wxWidgets returns logical pixels, but `sf::Mouse::setPosition` expects physical pixels](https://github.com/sirjuddington/SLADE/blob/fb7cc161a6886c12c5120e30190a7813c6c16b62/src/MapEditor/UI/MapCanvas.cpp#L119-L121).

This still worked because `MapCanvas::mouseLook3d()` later compensated by scaling `wxGetMousePosition()` from logical to physical pixels.

After commit 2a5e2b25e31e ("Use wxWidgets to center the mouse on the 3D map editor"), `MapCanvas::mouseToCenter` actually centers the mouse on scaled displays.

Now the compensatory logic causes the camera to spin uncontrollably, so remove it. All intermediate calculations now work in logical pixels, we scale to physical deltas at the end for consistent camera movement.

### DIFF
--- a/src/MapEditor/UI/MapCanvas.cpp
+++ b/src/MapEditor/UI/MapCanvas.cpp
@@ -157,24 +157,20 @@ void MapCanvas::mouseLook3d()
 		auto overlay_current = context_->currentOverlay();
 		if (!overlay_current || !overlay_current->isActive() || (overlay_current && overlay_current->allow3dMlook()))
 		{
-			// Get relative mouse movement (scale with dpi on macOS and Linux)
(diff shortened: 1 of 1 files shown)


joanbm:
(A sample app I used to test this: [pointer_test.cpp](https://github.com/user-attachments/files/23440971/pointer_test.cpp))

PR #1953, opened 2026-02-18
Build and use wxWidgets 3.3.1 for Linux AppImage CI build

sirjuddington:


Blzut3:
In case this is related to the Wayland AppImage request, do note that you probably need to build the latest GTK 3 release from source as well.

sirjuddington:
> In case this is related to the Wayland AppImage request, do note that you probably need to build the latest GTK 3 release from source as well.

Hmm we'll see when we get there :P Currently just trying to get linuxdeploy to find the wx libs. When I build locally using self-built wx3.3 it's fine just installing the libgtk-3-dev package though

Blzut3:
Edit: Nevermind, I confused the two issues.  You might be fine actually. Assuming the GTK issue was specific to the start page.

PR #1957, opened 2026-02-21
Make logo svg size be 512x512 instead of 64x64

electricbrass:
On Linux, it would sometimes be rasterized at 64x64 and then scaled up instead of being scaled up and then rasterized, resulting in blurry icons.


sirjuddington:
Hah, I'd just figured this out too and was about to commit a fix for it in 3.3.0 (along with a bunch of other linux-related changes).

This will fix it for 3.2.x in the meantime though, so I'll merge it in

PR #2006, opened 2026-06-07
Basic support for 3ds Max model format

wolfmanfp:
The HROT.pak file uses Quake II and 3ds Max model files, so I added some basic support for the latter.

### DIFF
--- a/dist/res/config/entry_types/model.json
+++ b/dist/res/config/entry_types/model.json
@@ -20,6 +20,11 @@
     "format": "mesh_md3",
     "category": "3d Models"
   },
+  "model_3ds": {
+    "name": "Model (3ds Max)",
+    "format": "mesh_3ds",
+    "category": "3d Models"
+  },
   // Also, voxels!
   "voxel_vox": {
     "name": "Voxel (VOX)",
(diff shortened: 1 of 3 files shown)


sirjuddington:
Looks good, however you'll need to re-implement this (and #2004/#2005) in the 3.3.0 branch if you can. I don't plan on making any more 3.2.x releases (master branch) and a lot of this has changed in 3.3.0. Sorry for the run around

wolfmanfp:
Np, I'll give it a try

wolfmanfp:
Done. Also added #2007, which was the remaining PR I wanted to create (and of course, based on the 3.3.0 branch).

sirjuddington:
Hmm, are you able to add more checking to the data format? Just checking for the first 2 bytes being 'MM' will likely lead to false positives

wolfmanfp:
The magic bytes are supposed to be just `MM`, so I don't know how to add any more checks than that.

sirjuddington:
Ok all good, I found a spec for the format so I should be able to add something extra (or at the very least require a .3ds extension)

PR #2007, opened 2026-06-07
Fix recognition of JPEG images without JFIF or Exif header

wolfmanfp:
There was this JPEG file in HROT.pak which was not recognised by SLADE, because it had no JFIF or Exif header. It turns out these are optional, and this is a valid JPEG file:

mesic

### DIFF
--- a/src/Archive/EntryType/DataFormats/ImageFormats.h
+++ b/src/Archive/EntryType/DataFormats/ImageFormats.h
@@ -253,13 +253,9 @@ public:
 		if (mc.size() > 128)
 		{
 			// Check for JPEG header
-			if ((mc[6] == 'J' && mc[7] == 'F' && mc[8] == 'I' && mc[9] == 'F')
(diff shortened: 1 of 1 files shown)


sirjuddington:
Similar to the 3ds one, removing the header checks without adding anything else will likely lead to false positives, not sure what to do about this one

wolfmanfp:
I've done a bit of research. Strictly speaking, the first two bytes should be enough as header, which is the Start Of Image marker of JPEG files. Most sources I found listed the first two or three bytes as magic number.  
For the sake of compliance, I added a check for the fourth byte as well, so it checks the whole first sequence tag. [This site](https://www.garykessler.net/library/file_sigs_GCK_latest.html) and [this GitHub comment](https://gist.github.com/leommoore/f9e57ba2aa4bf197ebc5?permalink_comment_id=3863054#gistcomment-3863054) explains it well, and helped me a lot. The latter also links to the specification.  
Another way would be checking the last two bytes (EOI - End Of Image -  `ff d9`), but from what I understand, that's not foolproof either.

PR #2008, opened 2026-06-07
Swap compiler output and compiler error messages

electricbrass:
When showing output from an ACS compiler, the two messages that can be shown at the top were swapped. This made it say there was an error when there wasn't and only say there was output when there was an error. When adding DECOHack support, I copied the ACS code and duplicated the bug there too.


electricbrass:
This same issue is present in 3.3.0, so will need to be fixed on that branch as well.

